PR #6, opened 2021-01-24
Add support for Jooq 3.11.x

effiban:
Needed for the **kea** microservice - to support certain Jooq operators that are not available before this version

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 project.ext.set("libraries", "")
 
 ext.jackson_version = "2.7.8"
-ext.jooq_version = project.hasProperty('JOOQ_VERSION') ? "$JOOQ_VERSION" : '3.10.4'
+ext.jooq_version = project.hasProperty('JOOQ_VERSION') ? "$JOOQ_VERSION" : '3.11.12'
 
 project.ext.libraries = [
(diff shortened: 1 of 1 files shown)


dependencies.gradle
```diff
@@ -1,7 +1,7 @@
 project.ext.set("libraries", "")
 
 ext.jackson_version = "2.7.8"
-ext.jooq_version = project.hasProperty('JOOQ_VERSION') ? "$JOOQ_VERSION" : '3.10.4'
+ext.jooq_version = project.hasProperty('JOOQ_VERSION') ? "$JOOQ_VERSION" : '3.11.12'
```
dimagorelik, line 4: @effiban 
Does it mean that version  3.10.4 will not be supported?

effiban: It is just the default version. I modified the release job to run on all 3 supported versions - 3.9.x., 3.10.x and now also 3.11.x

dimagorelik: @effiban 
We should think about it once more if we need to release PL per Jooq version.

effiban: @dimagorelik I agree but right now we need it and I don't want to delay the Titans team

effiban: @amirz @idopreuss @yardenadam - FYI

dimagorelik (approved):

PR #11, opened 2021-02-15
add postFetchCondition to PlCondition

SaraMunk:
SEARCH-95928
The UniquenessValidator checks duplication in bulk command and DB by a unique key and condition.
the validator check the condition just in the DB entities and did not check the condition on the command entities.
This PR add a predicate condition to the PLCondition beside the jooq condition, and the UniquenessValidator use this predicate condition on the bulk commands.
@galkoren

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -216,6 +216,28 @@ public class UniquenessValidatorTest {
     }
 
     @Test
+    public void testDontFailWhenAllCommandsUnmatchedCondition() {
+        final var validator = new UniquenessValidator
+                .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
+                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
+                .build();
+
+        final var entity1 = new CreateParent()
+                .with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, null);
+
+        final var entity2 = new CreateParent()
+                .with(ParentEntity.ID, 1)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, null);
+
+        final var results = parentPersistence.create(List.of(entity1, entity2), parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(false));
+    }
+
+    @Test
     public void testDontFailCommandWhenSameUniqueKeyInDBButConditionIsUnmatched() {
         UniqueKey<ParentEntity> uniqueness = new UniqueKey<>(asList(ParentEntity.NAME));
 
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -172,7 +172,7 @@ public class UniquenessValidatorTest {
     }
 
     @Test
-    public void testDontFailBulkCommandsWhenConditionIsUnmatched() {
+    public void testDontFailCommandWhenSameUniqueKeyInBulkButConditionIsUnmatched() {
         final var validator = new UniquenessValidator
                 .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
                 .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
@@ -194,7 +194,7 @@ public class UniquenessValidatorTest {
     }
 
     @Test
-    public void testFailBulkCommandsWhenConditionIsMatched() {
+    public void testFailCommandWhenSameUniqueKeyInBulkAndConditionIsMatched() {
         final var validator = new UniquenessValidator
                 .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
                 .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
@@ -216,7 +216,7 @@ public class UniquenessValidatorTest {
     }
 
     @Test
-    public void testDontFailCommandConditionIsUnmatched() {
+    public void testDontFailCommandWhenSameUniqueKeyInDBButConditionIsUnmatched() {
         UniqueKey<ParentEntity> uniqueness = new UniqueKey<>(asList(ParentEntity.NAME));
 
         UniquenessValidator<ParentEntity> validator = new UniquenessValidator
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -221,13 +221,17 @@ public class UniquenessValidatorTest {
 
         UniquenessValidator<ParentEntity> validator = new UniquenessValidator
                 .Builder<>(entitiesFetcher, uniqueness)
-                .setCondition(PLCondition.not(ParentEntity.NAME.eq("moshe")))
+                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
                 .build();
 
-        create(new CreateParent().with(ParentEntity.ID, 99).with(ParentEntity.NAME, "moshe"));
+        create(new CreateParent().with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, null));
 
         List<CreateParent> commands = asList(
-                new CreateParent().with(ParentEntity.ID, 1).with(ParentEntity.NAME, "moshe")
+                new CreateParent().with(ParentEntity.ID, 1)
+                        .with(ParentEntity.NAME, "moshe")
+                        .with(ParentEntity.ID_IN_TARGET, 12345)
         );
 
         CreateResult<ParentEntity, Identifier<ParentEntity>> results = parentPersistence.create(commands, parentFlow(validator).build());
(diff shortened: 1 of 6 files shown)


main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -66,24 +75,27 @@ public boolean equals(Object o) {
...
+                .isEquals();
     }
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder(17, 37)
-            .append(jooqCondition)
-            .append(fields)
-            .toHashCode();
+        return new HashCodeBuilder(17, 37)//??
```
SaraMunk, line 86: I will remove this comment in the next commit.but what are these numbers for? do I have to change it now once the lambdaCondition was added ?

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -8,53 +8,62 @@
...
-import java.util.Collections;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import static java.util.Objects.requireNonNull;
 
 public class PLCondition {
 
     private final Condition jooqCondition;
+    private final Predicate<Entity> lambdaCondition;
```
galkoren, line 19: lambdaCondition --> postFetchCondition

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -51,18 +41,21 @@ public void validate(Collection<? extends EntityChange<E>> commands, ChangeOpera
 
         Map<Identifier<E>, ? extends EntityChange<E>> commandsByIds = markDuplicatesInCollectionWithErrors(commands, ctx);
 
-        UniqueKey<E> pk = uniqueKey.getEntityType().getPrimaryKey();
-        EntityField<E, ?>[] uniqueKeyAndPK = ArrayUtils.addAll(uniqueKey.getFields(), pk.getFields());
+        if (!commandsByIds.isEmpty()) {
```
SaraMunk, line 44: in case no command to check duplication for in DB do not continue ( if so, it will fail in the EntitiesFetcher  for empty keys)

SaraMunk:
@chayamor FYI

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
```diff
@@ -171,19 +171,67 @@ public void testFailWhenAllPartsOfTheUniqueKeyMatched() {
         assertTrue(results.hasErrors());
     }
 
+    @Test
+    public void testDontFailBulkCommandsWhenConditionIsUnmatched() {
```
galkoren, line 175: Now we support condition in the bulk. Before we only supported condition in DB.
The name of the test should reflect that.

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -66,24 +75,27 @@ public boolean equals(Object o) {
         PLCondition that = (PLCondition) o;
 
         return new EqualsBuilder()
-            .append(jooqCondition, that.jooqCondition)
-            .append(fields, that.fields)
-            .isEquals();
+                .append(jooqCondition, that.jooqCondition)
+                .append(lambdaCondition, that.lambdaCondition)
```
galkoren, line 79: Don't add the lambdaCondition because it doesn't have a state (Java fields), it is a reference to code.
Adding a reference (pointer) value to equality means that two objects will never equal.
Test this:
```java
var c1 = FIELD1.eq(5);
var c2 = FIELD1.eq(5); // same as c1
assertThat(c1, equals(c2));  // will fail

```

galkoren: Actually, let's delete `equals` and `hashCode` completly. They can never work as expected.
```java
var c1 = COND_A.and(COND_B);
var c2 = COND_B.and(COND_A);
```
c1 should equal c2, right?
But it won't work

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
galkoren: I think PL skips empty collections before calling validators so it's not important

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
SaraMunk: yes, this is why the test name is testDontFail**Bulk**CommandsWhenConditionIsUnmatched..
but OK, I will change it to more clear name

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
SaraMunk: :+1:

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
SaraMunk: right, but in case all the commands will filtered by the postFetchCondition we will have an empty list and it will fail in the EntitiesFetcher that requires non empty keys.

SaraMunk:
@galkoren Can u please review again? thanks!

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -41,7 +41,7 @@ default PLCondition in(T ...values) {
         final Object[] tableValues = Arrays.stream(values).map(value -> getDbAdapter().getFirstDbValue(value)).toArray(Object[]::new);
         @SuppressWarnings("unchecked")
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        return new PLCondition(tableField.in(tableValues), this);
+        return new PLCondition(tableField.in(tableValues), entity -> Arrays.stream(values).anyMatch(value -> entity.safeGet(this).equals(Triptional.of(value))), this);
```
galkoren, line 44: Improve run time complexity to better than O(n*n)

SaraMunk: now the run time is O(n) as the values length.
so I will change it to List.contains - I think this is run time of O(1).

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -41,7 +42,9 @@ default PLCondition in(T ...values) {
         final Object[] tableValues = Arrays.stream(values).map(value -> getDbAdapter().getFirstDbValue(value)).toArray(Object[]::new);
         @SuppressWarnings("unchecked")
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        return new PLCondition(tableField.in(tableValues), this);
+        final var setOfValues = Seq.of(values).toSet();
```
SaraMunk, line 45: .toSet()  does not fail when duplication in values or  null value vs Set.of() that throws exception for duplication elements and throws exception for null element.

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -50,6 +53,6 @@ default PLCondition isNull() {
         }
         @SuppressWarnings("unchecked")
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        return new PLCondition(tableField.isNull(), this);
+        return new PLCondition(tableField.isNull(), entity -> entity.safeGet(this).isNull(), this);
```
SaraMunk, line 56: maybe better isNullOrAbsent?

galkoren (approved):

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -8,82 +8,70 @@
...
+                               condition.postFetchCondition.negate(),
                                condition.fields);
     }
 
     public static PLCondition trueCondition() {
-        return  new PLCondition(DSL.trueCondition());
-    }
-
-    @Override
-    public boolean equals(Object o) {
```
dimagorelik, line 61: @SaraMunk 
Why have you removed equals implementation?

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -8,82 +8,70 @@
...
-
-        return new EqualsBuilder()
-            .append(jooqCondition, that.jooqCondition)
-            .append(fields, that.fields)
-            .isEquals();
-    }
-
-    @Override
-    public int hashCode() {
-        return new HashCodeBuilder(17, 37)
```
dimagorelik, line 76: @SaraMunk 
The same question

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
SaraMunk: @dimagorelik 
see @galkoren comment that showed with example that these methods can never work as expected.

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -41,7 +42,9 @@ default PLCondition in(T ...values) {
         final Object[] tableValues = Arrays.stream(values).map(value -> getDbAdapter().getFirstDbValue(value)).toArray(Object[]::new);
         @SuppressWarnings("unchecked")
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        return new PLCondition(tableField.in(tableValues), this);
+        final var setOfValues = Seq.of(values).toSet();
+        return new PLCondition(tableField.in(tableValues),
+                entity -> setOfValues.contains(entity.safeGet(this).asOptional().orElse(null)), this);
```
effiban, line 47: This can be simplified:
A slightly 'cleaner' way to do this (if you also consider that `null` is invalid as above):
```
entity -> entity.safeGet(this)
                .filter(Triptional::isNotNull)
                .matches(setOfValues::contains)
```

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
effiban: `isNull()` is correct, since we want to know what is in the entity.
"Absent" means that the field does not appear at all, so we cannot know what its value is (maybe we didn't fetch it from the DB)

effiban (commented):
Sorry for the late review - please take a look - thanks.
If you don't have time, maybe someone else can do the fixes.

PR #12, opened 2021-02-22
PL-103: Support private entity fields for Scala

effiban:
PL uses reflection to resolve annotations on the entity fields, and also to give names to the entity fields for logging (plus another usage for fetching partial entities which is less important and may be deprecated soon).
This reflection uses  `getField(...)` and `field.get(null)`  which together assume that the fields are `public static`. This sets a requirement for how the fields must be defined, and it makes sense since the `EntityType` should be a singleton.

However Scala does not have an explicit `public` **nor** `static` for fields. Instead:
1) Fields are defined as `val` (by the best practice) which makes them immutable, and they are also `public` by default. However under the hood they are compiled as `private final` Java members and the `public` getters are generated synthetically.
2) Scala does not use `static`. The equivalent is a member inside a singleton, and this can be achieved either by placing it inside an `object` or using a private constructor like Java.

Due to the above, PL `EntityType`-s cannot be defined in Scala with the current reflection logic.
After thinking about this, my conclusion is that the simplest solution is to relax the requirement of `public static` for the fields and support `private` and non-`static` as well by using the `getDeclaredField(...)` reflection method.
The drawbacks as far as I see are:
1) PL no longer enforces a singleton `EntityType` in Java, since the fields could be defined as non-static and the `EntityType` instantiated multiple times. I think it's a minor risk and we can enforce it as a best practice with the Wiki and through the plugin.
2) The fields could be defined in Java as `private` with additional getters/setters that will not be scanned or recognized by the framework, and the user might try to place annotations on them which will not be recognized. Again this can be enforced with the Wiki/plugin.
3) `getField(..)` can also fetch `public` fields recursively up the hierarchy - which means we will no longer support inheritance of `EntityType`-s, but as far as I can tell nobody uses it (and I'm not sure it would work in all scenarios anyway).

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
@@ -8,6 +8,7 @@ import com.kenshoo.pl.entity.*;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
@@ -93,7 +94,9 @@ public abstract class EntityTypeReflectionUtil {
 
     private static <E extends EntityType<E>, T> EntityField<E, T> getEntityFieldInstance(final Field field, final EntityType<E> entityType) {
         try {
-            field.setAccessible(true);
+            if (!Modifier.isPublic(field.getModifiers())) {
+                field.setAccessible(true);
+            }
             //noinspection unchecked
             return (EntityField<E, T>) field.get(entityType);
         } catch (IllegalAccessException e) {
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
@@ -81,22 +81,21 @@ public abstract class EntityTypeReflectionUtil {
         }
     }
 
-    public static <E extends EntityType<E>> BiMap<String, EntityField<E, ?>> getFieldToNameBiMap(Class<? extends AbstractEntityType> entityClass) {
-        Map<String, EntityField<E, ?>> map = Stream.of(entityClass.getFields())
-                .filter(field -> EntityField.class.isAssignableFrom(field.getType()))
-                .collect(Collectors.toMap(
-                                Field::getName,
-                                EntityTypeReflectionUtil::getEntityFieldInstance
-                        )
-                );
+    public static <E extends EntityType<E>> BiMap<String, EntityField<E, ?>> getFieldToNameBiMap(final EntityType<E> entityType) {
+        Map<String, EntityField<E, ?>> map =
+            Stream.of(entityType.getClass().getDeclaredFields())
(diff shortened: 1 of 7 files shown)


effiban:
@amirz @katrinp1 - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
```diff
@@ -81,22 +81,21 @@ private static String toEntityFieldName(String methodName) {
...
+                  .collect(Collectors.toMap(Field::getName,
+                                            field -> getEntityFieldInstance(field, entityType)));
 
         return HashBiMap.create(map);
     }
 
-    private static <E extends EntityType<E>, T> EntityField<E, T> getEntityFieldInstance(Field field) {
+    private static <E extends EntityType<E>, T> EntityField<E, T> getEntityFieldInstance(final Field field, final EntityType<E> entityType) {
         try {
+            field.setAccessible(true);
```
dimagorelik, line 96: @effiban 
What does it mean?

effiban: This is how you can access a `private` field in reflection - it's not accessible by default.

main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
```diff
@@ -81,22 +81,21 @@ private static String toEntityFieldName(String methodName) {
...
-        Map<String, EntityField<E, ?>> map = Stream.of(entityClass.getFields())
-                .filter(field -> EntityField.class.isAssignableFrom(field.getType()))
-                .collect(Collectors.toMap(
-                                Field::getName,
-                                EntityTypeReflectionUtil::getEntityFieldInstance
-                        )
-                );
+    public static <E extends EntityType<E>> BiMap<String, EntityField<E, ?>> getFieldToNameBiMap(final EntityType<E> entityType) {
+        Map<String, EntityField<E, ?>> map =
+            Stream.of(entityType.getClass().getDeclaredFields())
```
dimagorelik, line 86: @effiban 
Why the method getFields() is replaced by getDeclaredFields()?

effiban: @dimagorelik see description above, I explained everything there.

main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
dimagorelik, line 96: @effiban 
Can we receive a security exception here?

effiban: I looked at the documentation. It seems that the only issue could be if the application has defined a `SecurityManager`, which is beyond our control in any case (and it's true for any reflection). See here:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/AccessibleObject.html#setAccessible(boolean)

dimagorelik: @effiban 
I have a very bad experience with such a solution in the past.

effiban:
retest this please

dimagorelik (approved):

PR #13, opened 2021-02-23
support update operation in UniquenessValidator

SaraMunk:
SEARCH-96112
support the UniquenessValidator to validate also update commands.
@galkoren Can u please review? thanks!
@chayamor FYI

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -261,6 +261,49 @@ public class UniquenessValidatorTest {
         assertFalse(results.hasErrors());
     }
 
+    @Test
+    public void testFailCommandWhenSameUniqueKeyInDBAndConditionIsMatched() {
+        final var validator = new UniquenessValidator
+                .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
+                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
+                .build();
+
+        create(new CreateParent().with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, 333));
+
+        final var commands = List.of(
+                new CreateParent().with(ParentEntity.ID, 1)
+                        .with(ParentEntity.NAME, "moshe")
+                        .with(ParentEntity.ID_IN_TARGET, 12345)
+        );
+
+        final var results = parentPersistence.create(commands, parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(true));
+    }
+
+    @Test
+    public void testDontFailUpdateCommandOnItselfInDBForSameUniqueKeyError() {
+        final var validator = new UniquenessValidator
+                .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
+                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
+                .setOperation(SupportedChangeOperation.CREATE_AND_UPDATE)
+                .build();
+
+        create(new CreateParent().with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, 333));
+
+        final var updateCommand =
+                new UpdateParent(99)
+                        .with(ParentEntity.NAME, "moshe")
+                        .with(ParentEntity.ID_IN_TARGET, 333);
+
+        final var results = parentPersistence.update(List.of(updateCommand), parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(false));
+    }
 
     private List<CreateEntityCommand<ParentEntity>> failedCommands(CreateResult<ParentEntity, Identifier<ParentEntity>> results) {
         return results.getChangeResults().stream().filter(not(r -> r.isSuccess())).map(r -> r.getCommand()).collect(toList());
@@ -276,6 +319,12 @@ public class UniquenessValidatorTest {
     private static class CreateParent extends CreateEntityCommand<ParentEntity> implements EntityCommandExt<ParentEntity, CreateParent> {
         public CreateParent() {
             super(ParentEntity.INSTANCE);
+        }
+    }
+
+    private static class UpdateParent extends UpdateEntityCommand<ParentEntity, ParentEntity.Key> implements EntityCommandExt<ParentEntity, UpdateParent> {
+        public UpdateParent(int id) {
+            super(ParentEntity.INSTANCE, new ParentEntity.Key(id));
         }
     }
 
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -284,23 +284,56 @@ public class UniquenessValidatorTest {
     }
 
     @Test
-    public void testDontFailUpdateCommandOnItselfInDBForSameUniqueKeyError() {
+    public void testDontFailUpdateCommandOnItselfInDBWhenNoChangeInUniqueKeyFields() {
         final var validator = new UniquenessValidator
                 .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
-                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
                 .setOperation(SupportedChangeOperation.CREATE_AND_UPDATE)
                 .build();
 
-        create(new CreateParent().with(ParentEntity.ID, 99)
-                .with(ParentEntity.NAME, "moshe")
-                .with(ParentEntity.ID_IN_TARGET, 333));
+        create(new CreateParent().with(ParentEntity.ID, 99).with(ParentEntity.NAME, "moshe"));
 
-        final var updateCommand =
-                new UpdateParent(99)
-                        .with(ParentEntity.NAME, "moshe")
-                        .with(ParentEntity.ID_IN_TARGET, 333);
+        final var updateCommand = new UpdateParent(99).with(ParentEntity.NAME, "moshe");
 
         final var results = parentPersistence.update(List.of(updateCommand), parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(false));
+    }
+
+    @Test
+    public void testNullInUniqueKeyIsConsideredAsValueForComparison() {
+        final var validator = new UniquenessValidator
+                .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
+                .build();
+
+        final var entity1 = new CreateParent()
+                .with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, null);
+
+        final var entity2 = new CreateParent()
+                .with(ParentEntity.ID, 1)
+                .with(ParentEntity.NAME, null);
+
+        final var entity3 = new CreateParent()
+                .with(ParentEntity.ID, 1)
+                .with(ParentEntity.NAME, "moshe");
+
+        final var results = parentPersistence.create(List.of(entity1, entity2, entity3), parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(true));
+        assertThat(results.getErrors(entity2).isEmpty(), is(false));
+        assertThat(results.getErrors(entity3).isEmpty(), is(true));
+    }
+
+    @Test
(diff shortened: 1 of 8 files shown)


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -44,19 +48,29 @@ public void validate(Collection<? extends EntityChange<E>> commands, ChangeOpera
         if (!commandsByIds.isEmpty()) {
             UniqueKey<E> pk = uniqueKey.getEntityType().getPrimaryKey();
             EntityField<E, ?>[] uniqueKeyAndPK = ArrayUtils.addAll(uniqueKey.getFields(), pk.getFields());
+            final var updateCommandsPKs = getPrimaryKeysOfUpdateCommands(ctx, commandsByIds, pk);
 
             Map<Identifier<E>, CurrentEntityState> duplicates = fetcher.fetch(uniqueKey.getEntityType(), commandsByIds.keySet(), condition, uniqueKeyAndPK)
-                    .stream().collect(toMap(e -> createKeyValue(e, uniqueKey), identity()));
+                    .stream()
+                    .filter(entity -> !updateCommandsPKs.contains(pk.createIdentifier(entity)))
```
SaraMunk, line 55: This filter prevent  update command to fail with duplication error for itself in DB.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -70,8 +84,8 @@ private ValidationError errorForDatabaseConflict(CurrentEntityState dupEntity, U
         };
     }
 
-    private static <E extends EntityType<E>> Identifier<E> createKeyValue(EntityChange<E> cmd, UniqueKey<E> key) {
-        return key.createIdentifier(cmd);
+    private static <E extends EntityType<E>> Identifier<E> createKeyValue(EntityChange<E> cmd, ChangeContext ctx, UniqueKey<E> key) {
+        return key.createIdentifier(ctx.getFinalEntity(cmd));
```
SaraMunk, line 88: In order to support update - need to build the unique key identifier by the final entity.( because in update in case unique key fields were not changed it will fail here for noValueSupplied so need to get the values from the db entity)

effiban: I'm not sure this is correct.
This method is called on line 73 while checking that **commands** do not clash, and it is not supposed to check against the DB because that is done in the next part. So if you are using the final entity here you are actually checking against the DB already and catching the duplicate too soon.
Instead you can simply filter out (before line 73) - all commands which do _not_ have the unique key.

This is not really a bug, but still it means that the wrong error code will appear in case of failure (so please check that as well in your tests).

SaraMunk: I can not filter the commands before because the unique key can be more than one field so can be a case that update command has value just for on unique key field and not for the other fields in the unique key so we need to check duplication for this command but we fill the other fields values from the entity in db.

effiban: ok, it's more complicated then I thought.. let's leave it as is :)

effiban (approved):

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
```diff
@@ -261,6 +261,49 @@ public void testDontFailCommandWhenSameUniqueKeyInDBButConditionIsUnmatched() {
...
+        final var results = parentPersistence.create(commands, parentFlow(validator).build());
+
+        assertThat(results.hasErrors(), is(true));
+    }
+
+    @Test
+    public void testDontFailUpdateCommandOnItselfInDBForSameUniqueKeyError() {
+        final var validator = new UniquenessValidator
+                .Builder<>(entitiesFetcher, new UniqueKey<>(List.of(ParentEntity.NAME)))
+                .setCondition(PLCondition.not(ParentEntity.ID_IN_TARGET.isNull()))
```
galkoren, line 290: why do we need ID_IN_TARGET in this test if the uniqueKey is NAME ?

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
```diff
@@ -261,6 +261,49 @@ public void testDontFailCommandWhenSameUniqueKeyInDBButConditionIsUnmatched() {
...
+                .setOperation(SupportedChangeOperation.CREATE_AND_UPDATE)
+                .build();
+
+        create(new CreateParent().with(ParentEntity.ID, 99)
+                .with(ParentEntity.NAME, "moshe")
+                .with(ParentEntity.ID_IN_TARGET, 333));
```
galkoren, line 296: same question

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
```diff
@@ -261,6 +261,49 @@ public void testDontFailCommandWhenSameUniqueKeyInDBButConditionIsUnmatched() {
...
+
+        final var updateCommand =
+                new UpdateParent(99)
+                        .with(ParentEntity.NAME, "moshe")
+                        .with(ParentEntity.ID_IN_TARGET, 333);
```
galkoren, line 301: same question

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -44,19 +48,29 @@ public void validate(Collection<? extends EntityChange<E>> commands, ChangeOpera
...
-                    .stream().collect(toMap(e -> createKeyValue(e, uniqueKey), identity()));
+                    .stream()
+                    .filter(entity -> !updateCommandsPKs.contains(pk.createIdentifier(entity)))
+                    .collect(toMap(e -> createKeyValue(e, uniqueKey), identity()));
 
             duplicates.forEach((dupKey, dupEntity) -> ctx.addValidationError(commandsByIds.get(dupKey), errorForDatabaseConflict(dupEntity, pk)));
         }
     }
 
+    private Set<Identifier<E>> getPrimaryKeysOfUpdateCommands(ChangeContext ctx, Map<Identifier<E>, ? extends EntityChange<E>> commandsByIds, UniqueKey<E> pk) {
```
galkoren, line 62: don't pass the `Map` if you don't need a map. Pass only the `map.values()`

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
galkoren: `createIdentifer` creates an array in memory and we don't need to call it for CREATE commands.
We can optimize the filter:
```java
e -> !(e.getOperation() == UPDATE && updateCommandsPKs.contains(pk.createIdentifier(entity))
```
and let's extract this filter to a method:
```java
  .stream()
  .filter(notUpdateIn(updateCommandsPKs))
```

effiban:
retest this please

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
SaraMunk: we filter here the entities from DB. not the commands.
the updateCommandsPKs has the PKs list just for the Update commands, and we filter here the entities from DB that exist in this updateCommandsPKs list.

integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
SaraMunk: in this test the ID_IN_TARGET is not necessary but for example in this test - testDontFailCommandWhenSameUniqueKeyInBulkButConditionIsUnmatched I need the condition to be different than the unique key (so the test wont fail because of the condition and not because no duplication..)
I can remove it in this test but I thought it is more clear to have the same validate in all tests.

dannielshalev:
retest this please

dannielshalev:
retest this please

dannielshalev:
retest this, please

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -42,9 +44,9 @@ default PLCondition in(T ...values) {
         final Object[] tableValues = Arrays.stream(values).map(value -> getDbAdapter().getFirstDbValue(value)).toArray(Object[]::new);
         @SuppressWarnings("unchecked")
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        final var setOfValues = Seq.of(values).toSet();
+        final var setOfValues = Seq.of(values).collect(Collectors.toUnmodifiableSet());
         return new PLCondition(tableField.in(tableValues),
-                entity -> setOfValues.contains(entity.safeGet(this).asOptional().orElse(null)), this);
+                entity -> entity.safeGet(this).filter(Objects::nonNull).matches(setOfValues::contains), this);
```
SaraMunk, line 49: need to filter null values to prevent nullPointerException in contains method.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
SaraMunk: @galkoren 
I removed this filter because this is redundant now once we filter the update commands that have no change in unique key field.(line 48)

main/src/main/java/com/kenshoo/pl/entity/EntityWithNullForMissingField.java
```diff
@@ -0,0 +1,20 @@
+package com.kenshoo.pl.entity;
+
+public class EntityWithNullForMissingField implements Entity {
```
dimagorelik, line 3: @SaraMunk 
I do not understand what is the purpose of this class?

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
dimagorelik: @SaraMunk 
I think that it is not correct to filter out nulls. It means that something is not correct in logic. Maybe should be thrown exception?

SaraMunk: The setOfValues can not have null values. in case on of the values is null a nullPointerException will be thrown in line 44 when building the jooqCondition.
so we are OK to filter null because the setOfValues can never contains nulls.

effiban (approved):
I checked the changes relevant to my comments, looks good

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
dimagorelik: @SaraMunk 
Let's remove this filter

main/src/main/java/com/kenshoo/pl/entity/EntityWithNullForMissingField.java
SaraMunk: As talked with @galkoren we need this class to pass it to the UniqueKey:createIdentifier method to prevent IllegalArgumentException of field not fetched in create commands that missing any unique key field.

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
effiban: @dimagorelik - why? it is not different than Jooq behavior

main/src/main/java/com/kenshoo/pl/entity/EntityWithNullForMissingField.java
dimagorelik: @SaraMunk 
I do not think that it is necessary, but if you need it, please do it as an inner private class.

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
dimagorelik: @effiban 
I do not see that Jooq filter out null values

effiban: @dimagorelik if you follow what Jooq and MySQL actually do - you will see that if there are any nulls in an 'IN' statement they are effectively ignored.
For example if I write this:
`WHERE field IN ('a', 'b', 'c', NULL)` and the field is `NULL` in the DB - the condition will be `false` and the `NULL` had no effect, because in SQL you have to use `IS NULL` in order to fetch `NULL` values.
So if we skip `null`-s here we are doing the same thing exactly.

dimagorelik: @effiban 
ok

galkoren (approved):

SaraMunk:
I run ./gradlew test  and it passed.

PR #18, opened 2021-03-08
add condition fields to requiredFields

SaraMunk:
SEARCH-95928
part of [this PR ](https://github.com/kenshoo/persistence-layer/pull/11) the uniquenessValidator check the condition on the bulk commands.
This PR add the condition fields to the requiredFields because when checking the condition on the bulk commands by the finalEntity these fields are necessary .
@galkoren

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
@@ -93,7 +93,8 @@ public class UniquenessValidator<E extends EntityType<E>> implements ChangesVali
 
     @Override
     public Stream<? extends EntityField<?, ?>> requiredFields(Collection<? extends EntityField<E, ?>> fieldsToUpdate, ChangeOperation op) {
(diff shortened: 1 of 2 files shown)


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -93,7 +93,8 @@ private ValidationError errorForDatabaseConflict(CurrentEntityState dupEntity, U
 
     @Override
     public Stream<? extends EntityField<?, ?>> requiredFields(Collection<? extends EntityField<E, ?>> fieldsToUpdate, ChangeOperation op) {
-        return Stream.concat(Stream.of(uniqueKey.getFields()), Stream.of(uniqueKey.getEntityType().getPrimaryKey().getFields()));
+        return Stream.of(condition.getFields().stream(), Stream.of(uniqueKey.getFields()), Stream.of(uniqueKey.getEntityType().getPrimaryKey().getFields()))
+                .flatMap(fields -> fields);
```
SaraMunk, line 97: @galkoren 
should we distinct it? ( because for example a uniqueKey field can be in a condition)

galkoren (approved):

PR #27, opened 2021-04-11
PL-113: In AuditRecord replace fields with their names

effiban:
In `AuditRecord` replacing all `EntityField` usages with strings representing the field names.

The fields are currently nested inside two other objects:
 - `EntityFieldValue` which is used to contain a mandatory field (field which always appears: "descriptive field", agency etc.) - and its value.
 The usages of this object have been completely replaced by a simple `Map.Entry<String, ?>`
 - `FieldAuditRecord` which is used to contain: (changed) field, old value and new value. Here I just replaced the `EntityField` with a `String`.
 
 The field names currently have a default which is the `toString()` of the name.
 In the next PR-s I will support the overrides.

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
@@ -1,32 +1,31 @@
 package com.kenshoo.pl.entity.audit;
 
 import com.kenshoo.pl.entity.ChangeOperation;
-import com.kenshoo.pl.entity.EntityFieldValue;
-import com.kenshoo.pl.entity.EntityType;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
 import java.util.Collection;
+import java.util.Map.Entry;
 
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
-public class AuditRecord<E extends EntityType<E>> {
+public class AuditRecord {
     private final String entityType;
     private final String entityId;
-    private final Collection<? extends EntityFieldValue> mandatoryFieldValues;
+    private final Collection<? extends Entry<String, ?>> mandatoryFieldValues;
     private final ChangeOperation operator;
-    private final Collection<? extends FieldAuditRecord<E>> fieldRecords;
-    private final Collection<? extends AuditRecord<?>> childRecords;
+    private final Collection<? extends FieldAuditRecord> fieldRecords;
+    private final Collection<? extends AuditRecord> childRecords;
 
     private AuditRecord(final String entityType,
                         final String entityId,
-                        final Collection<? extends EntityFieldValue> mandatoryFieldValues,
+                        final Collection<? extends Entry<String, ?>> mandatoryFieldValues,
                         final ChangeOperation operator,
-                        final Collection<? extends FieldAuditRecord<E>> fieldRecords,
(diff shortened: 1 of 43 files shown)


main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
```diff
@@ -36,10 +36,10 @@
...
     }
 
-    private <E extends EntityType<E>> FieldAuditRecord<E> buildFieldRecord(final CurrentEntityState currentState,
-                                                                           final FinalEntityState finalState,
-                                                                           final EntityField<E, ?> field) {
-        final FieldAuditRecord.Builder<E> fieldRecordBuilder = FieldAuditRecord.builder(field);
+    private <E extends EntityType<E>> FieldAuditRecord buildFieldRecord(final CurrentEntityState currentState,
+                                                                        final FinalEntityState finalState,
+                                                                        final EntityField<E, ?> field) {
+        final FieldAuditRecord.Builder fieldRecordBuilder = FieldAuditRecord.builder(field.toString());
```
effiban, line 42: Default field name is the `toString()`

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditMandatoryFieldValuesGenerator.java
```diff
@@ -21,12 +22,12 @@ public AuditMandatoryFieldValuesGenerator(final Stream<? extends EntityField<?,
...
     }
 
-    Collection<EntityFieldValue> generate(final FinalEntityState finalState) {
+    Collection<Entry<String, ?>> generate(final FinalEntityState finalState) {
         requireNonNull(finalState, "finalState is required");
         return seq(mandatoryFields)
             .map(field -> ImmutablePair.of(field, finalState.safeGet(field)))
             .filter(pair -> pair.getValue().isNotNull())
-            .map(pair -> new EntityFieldValue(pair.getKey(), pair.getValue().get()))
+            .map(pair -> entry(pair.getKey().toString(), pair.getValue().get()))
```
effiban, line 30: Default field name is `toString()`, and will remain the only possible value since we have currently no way to override this by annotation (we can discuss another way).

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
```diff
@@ -1,32 +1,31 @@
...
 
 import java.util.Collection;
+import java.util.Map.Entry;
 
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
-public class AuditRecord<E extends EntityType<E>> {
+public class AuditRecord {
```
effiban, line 15: Once fields are replaced with names, there is no need for this generic param anymore (same for `FieldAudittRecord`)

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
```diff
@@ -1,32 +1,31 @@
...
     private final ChangeOperation operator;
-    private final Collection<? extends FieldAuditRecord<E>> fieldRecords;
-    private final Collection<? extends AuditRecord<?>> childRecords;
+    private final Collection<? extends FieldAuditRecord> fieldRecords;
+    private final Collection<? extends AuditRecord> childRecords;
 
     private AuditRecord(final String entityType,
                         final String entityId,
-                        final Collection<? extends EntityFieldValue> mandatoryFieldValues,
+                        final Collection<? extends Entry<String, ?>> mandatoryFieldValues,
```
effiban, line 25: Restore to a FieldValue class

galkoren (approved):

PR #32, opened 2021-04-19
PL-119: In AuditRecord replace Entry<String, String> with FieldValue class

effiban:
Added a new `FieldValue` class which contains two strings and used it in `AuditRecord` instead of `Entry<String, String>`


main/src/main/java/com/kenshoo/pl/entity/internal/EntityFieldValue.java
```diff
@@ -1,5 +1,6 @@
-package com.kenshoo.pl.entity;
+package com.kenshoo.pl.entity.internal;
```
effiban, line 1: This class (which has a typed field and value as opposed to strings) is used only for some tests unrelated to auditing and also does not need to be exposed to the clients, so I moved it to the `internal` package

dimagorelik (approved):

PR #34, opened 2021-05-18
PL-123: Add the ability to disable auditing for a specific flow

effiban:
If there is no `DbCommandsOutputGenerator` in the flow (as in simulation mode), then no DB changes will occur and therefore it is likely that the client will not want to perform auditing either.
To support this:
1) Added methods to the flow builder to enable/disable auditing.
2) Auditing is disabled whenever `withoutOutputGenerators()` is called (this is to make migration easier in Kenshoo. Later on we can separate the two calls)

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
@@ -142,6 +142,7 @@ public class ChangeFlowConfig<E extends EntityType<E>> {
         private final List<ChangeFlowConfig.Builder<? extends EntityType<?>>> flowConfigBuilders = new ArrayList<>();
         private PersistenceLayerRetryer retryer = JUST_RUN_WITHOUT_CHECKING_DEADLOCKS;
         private final AuditedEntityTypeResolver auditedEntityTypeResolver;
+        private boolean auditingEnabled = true;
         private FeatureSet features = FeatureSet.EMPTY;
 
         public Builder(E entityType) {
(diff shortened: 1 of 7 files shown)


main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -280,11 +280,13 @@ public Builder(E entityType) {
             validators.forEach(validator -> validatorList.add(validator.element()));
             falseUpdatesPurger.ifPresent(enrichers::add);
 
-            final Optional<AuditedEntityType<E>> optionalAuditedEntityType = auditedEntityTypeResolver.resolve(entityType);
-            final AuditRequiredFieldsCalculator<E> auditRequiredFieldsCalculator =
-                optionalAuditedEntityType.map(AuditRequiredFieldsCalculator::new).orElse(null);
-            final AuditRecordGenerator<E> auditRecordGenerator =
-                optionalAuditedEntityType.map(this::createAuditRecordGenerator).orElse(null);
+            boolean hasDbOutputGenerator = outputGenerators.stream()
```
dimagorelik, line 283: @effiban 
I think we can disable auditing in witoutOtputGenerator method

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -248,6 +249,11 @@ public Builder(E entityType) {
             return this;
         }
 
+        public Builder<E> disableAuditing() {
```
dimagorelik, line 252: @effiban 
Maybe we need to add an option to enable auditing back?

effiban: @dimagorelik - I started out like that, with a method that receives a boolean - but I think this is simpler and less confusing.
If we want to disable both output generators and auditing together, I think we should add a separate method for doing both.

dimagorelik: @effiban 
In another hand, we can add only the disableAuditing, but we need to fix the code in KS in all places that apply withoutOtputGenerataors
.

effiban: I thought there weren't so many - doesn't seem like a big refactor

dimagorelik (approved):

PR #35, opened 2021-05-19
use thread local tx in makeChanges

galkoren:
TL;DR - the tx workaround was not working as expected.

Before this fix, when trying to use thread local tx lfailed with an exception:
```java
void myTest() {
    jooq.transaction(() -> pl.doSomething(...));
}
```
This test opens a tx on the local thread while PL is creating a different transaction on a derived DSL context.

As a workaround, we did something that did not throw an exception, but was wrong because PL still created a separate tx on a separate connection.
```java
var txProvider = jooq.configuration().transactionProvider();
var trx = new TransactionContextImpl(jooq.configuration(), jooq);
txProvider.begin(trx);

pl.doSomething(...);

txProvider.end(trx);
```

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/transaction/MultipleEntityTypeTransactionTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/transaction/MultipleEntityTypeTransactionTest.java
@@ -3,23 +3,23 @@ package com.kenshoo.pl.transaction;
 import com.kenshoo.jooq.DataTable;
 import com.kenshoo.jooq.DataTableUtils;
 import com.kenshoo.jooq.TestJooqConfig;
-import com.kenshoo.jooq.TransactionContextImpl;
 import com.kenshoo.pl.TypedFluidPersistenceCmdBuilder;
 import com.kenshoo.pl.entity.*;
 import com.kenshoo.pl.one2many.relatedByPK.IdGenerator;
 import com.kenshoo.pl.one2many.relatedByPK.IntegerIdGeneratorEnricher;
 import org.jooq.DSLContext;
 import org.jooq.Query;
-import org.jooq.TransactionContext;
-import org.jooq.TransactionProvider;
 import org.jooq.impl.DSL;
 import org.jooq.lambda.Seq;
-import org.junit.*;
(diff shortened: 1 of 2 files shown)


main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -123,7 +123,7 @@ private ChangeContext makeChanges(Collection<? extends ChangeEntityCommand<ROOT>
         Collection<? extends ChangeEntityCommand<ROOT>> validCmds = seq(commands).filter(cmd -> !context.containsError(cmd)).toList();
         ChangeContext overridingCtx = new OverridingContext(context);
         if (!validCmds.isEmpty()) {
-            flowConfig.retryer().run((() -> dslContext().transaction((configuration) -> generateOutputRecursive(flowConfig, validCmds, overridingCtx))));
+            flowConfig.retryer().run((() -> dslContext().transaction(() -> generateOutputRecursive(flowConfig, validCmds, overridingCtx))));
```
galkoren, line 126: Before, we used a derived config.
If we really wanted to use the tx opened by the caller, we should have passed the TransactionContext of the caller into here and create a new DSL from this context.

integration-tests/src/test/java/com/kenshoo/pl/transaction/MultipleEntityTypeTransactionTest.java
```diff
@@ -59,50 +59,39 @@ public void setupTables() {
...
     @Test
     public void rollbackTwoUnrelatedTypesInTransaction() {
 
-
         final var result1 = insertEntity1(newEntity1().with(Entity1.NAME, "name1"));
         final var result2 = insertEntity2(newEntity2().with(Entity2.NAME, "name2"));
         final var results =  List.of(result1.getChangeResults().iterator().next().getCommand().get(Entity1.ID),
                 result2.getChangeResults().iterator().next().getCommand().get(Entity2.ID));
 
-        final TransactionProvider txProvider = jooq.configuration().transactionProvider();
```
dimagorelik, line 89: @galkoren 
Does the removed implementation also work?

PR #39, opened 2021-06-22
Try to rework unique validator to make it possible to validate count …

AUsachov:
…of entities


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -29,47 +31,61 @@
...
     public void validate(Collection<? extends EntityChange<E>> commands, ChangeOperation op, ChangeContext ctx) {
 
         final var commandsToValidate = commands.stream().filter(hasChangeInUniqueKeyField()).collect(Collectors.toList());
-        Map<Identifier<E>, ? extends EntityChange<E>> commandsByIds = markDuplicatesInCollectionWithErrors(commandsToValidate, ctx);
-        if (commandsByIds.isEmpty())
+        var groupedCommands = groupByUniqueKey(commandsToValidate, ctx);
+
+        groupedCommands.forEach((eIdentifier, entityChanges) ->
+                entityChanges.stream()
+                        .skip(maxCount)
```
galkoren, line 59: `skip` is a clever idea :)
but lets not add errors yet because we also add them later, and we don't need to add the same error twice.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -29,47 +31,61 @@
...
+                        .forEach(eEntityChange -> ctx.addValidationError(eEntityChange, new ValidationError(errorCode))));
+        if (groupedCommands.isEmpty())
             return;
 
         UniqueKey<E> pk = uniqueKey.getEntityType().getPrimaryKey();
         EntityField<E, ?>[] uniqueKeyAndPK = ArrayUtils.addAll(uniqueKey.getFields(), pk.getFields());
 
-        Map<Identifier<E>, CurrentEntityState> duplicates = fetcher.fetch(uniqueKey.getEntityType(), commandsByIds.keySet(), condition, uniqueKeyAndPK)
+        Map<Identifier<E>, List<CurrentEntityState>> duplicates = fetcher.fetch(uniqueKey.getEntityType(), groupedCommands.keySet(), condition, uniqueKeyAndPK)
```
galkoren, line 67: "duplicates" is not a good name anymore. 
we can call them "itemsInDB"

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
```diff
@@ -29,47 +31,61 @@
...
                 .stream()
-                .collect(toMap(e -> createKeyValue(e, uniqueKey), identity()));
-
-        duplicates.forEach((dupKey, dupEntity) -> ctx.addValidationError(commandsByIds.get(dupKey), errorForDatabaseConflict(dupEntity, pk)));
+                .collect(groupingBy(e -> createKeyValue(e, uniqueKey), toList()));
+
+        duplicates.forEach((eIdentifier, currentEntityStates) -> {
+            List<? extends EntityChange<E>> identifierCommands = groupedCommands.get(eIdentifier);
```
galkoren, line 72: we can delete the "in same bulk" detection (line 57) and have a single check here:
```java
groupedCommands.forEach( (group, cmds) -> {
    int maxAllowedInBulk = maxCount - itemsFromDB.get(group).size() ;
    // now do the "skip" trick with maxAllowedInBulk
};
```

AUsachov: we'll get wrong error messages. I don't know is it critical, but when we have duplicate in db existed code should pass db id of duplicate

AUsachov: One more interesting question - what should be in error message when we have 3 items in db, maxCount is 3 and few items in commands list?

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidator.java
AUsachov: i think we should delete it from list, like it was before. As i understand in existed logic on first stage we check duplicates in commands, remove it and that compare with db what's left

PR #40, opened 2021-06-27
get first unique row when duplicates already exists in the db

chayamor:
When validate unique entity , and the entity already duplicate and not unique in the db - we failed on :

(attempted merging values {ADGROUP_ID=1} and {ADGROUP_ID=1})
	at java.base/java.util.stream.Collectors.duplicateKeyException

On this case - it is ok just to take the first entity from the db 

@galkoren @dimagorelik  Please review.

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
@@ -344,12 +344,11 @@ public class UniquenessValidatorTest {
                 .build();
 
         create(new CreateParent().with(ParentEntity.ID, 99)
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/com/kenshoo/pl/entity/spi/helpers/UniquenessValidatorTest.java
```diff
@@ -333,6 +336,33 @@ public void testFailCommandWhenSameUniqueKeyInDBAndConditionIsMatched() {
         assertThat(results.hasErrors(), is(true));
     }
 
+    @Test
+    public void testFailCommandWhenSameUniqueKeyInDBExistsMoreAndAlreadyDuplicateAndConditionIsMatched() {
```
chayamor, line 340: @galkoren  This test failed before the fix

galkoren (approved):

PR #42, opened 2021-06-27
create max count validator

galkoren:


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
```diff
@@ -0,0 +1,119 @@
...
+        this.operation = requireNonNull(operation, "operation must be provided");
+        if (maxAllowed < 1) {
+            throw new IllegalArgumentException("maxAllowed must be greater than zero");
+        }
+        this.maxAllowed = maxAllowed;
+    }
+
+    @Override
+    public SupportedChangeOperation getSupportedChangeOperation() {
+        return operation;
```
dimagorelik, line 43: @galkoren 
Make sense that this validation will be applied only for CREATE operation

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Override
+    public void validate(Collection<? extends EntityChange<E>> commands, ChangeOperation op, ChangeContext ctx) {
+
+        final var commandsToValidate = seq(commands)
+                .filter(command -> condition.getPostFetchCondition().test(ctx.getFinalEntity(command)))
```
dimagorelik, line 50: @galkoren 
Please, explain for which use case it should be used?

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -92,6 +94,33 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
...
+        final var aliasedKey = new AliasedKey<>(uniqueKey);
+        final var allFieldsToFetch = Seq.concat(seq(plCondition.getFields()), Stream.of(uniqueKey.getFields())).distinct().toList();
+        final var executionPlan = new ExecutionPlan(entityType.getPrimaryTable(), allFieldsToFetch).getOneToOnePlan();
+        final Field<Integer> countField = count();
+
+        final var queryBuilder = new QueryBuilder<E>(dslContext)
+                .selecting(dbFieldsOf(allFieldsToFetch).concat(countField).concat(aliasedKey.aliasedFields()).toList())
+                .from(entityType.getPrimaryTable())
+                .innerJoin(executionPlan.getPaths())
+                .leftJoin(executionPlan.getSecondaryTableRelations())
```
dimagorelik, line 114: @galkoren 
Maybe it should right join?

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
```diff
@@ -0,0 +1,119 @@
...
+            var numOfValidCommands = maxAllowed - countsInDB.getOrDefault(groupId, 0);
+            seq(changes).skip(numOfValidCommands).forEach(cmd -> markFailure(cmd, ctx));
+        });
+    }
+
+    private void markFailure(EntityChange<E> cmd, ChangeContext ctx) {
+        ctx.addValidationError(cmd, new ValidationError(errorCode));
+    }
+
+    private Predicate<EntityChange<E>> hasChangeInUniqueKeyField() {
```
dimagorelik, line 67: @galkoren 
I do not understand this predicate

effiban (commented):
In general LGTM, I have one question below.

Also a general comment:
I see that this validator is for the following (correct me if I'm wrong):
Given some unique-key combination, make sure that after all commands have been executed - no more than N rows exist in the table altogether.

This is fine, but just by looking at the name it's hard to get that. The user might only care about the total number of unique commands, regardless of what's in the DB.
I think we had this discussion regarding the `UniquenessValidator` too - might be worth using a more specific name, or else add another flag indicating whether the DB should be considered.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
```diff
@@ -0,0 +1,119 @@
...
+        return cmd -> !UPDATE.equals(cmd.getChangeOperation()) || Arrays.stream(groupingKey.getFields()).anyMatch(cmd::isFieldChanged);
+    }
+
+    private static <E extends EntityType<E>> Identifier<E> createKeyValue(EntityChange<E> cmd, ChangeContext ctx, UniqueKey<E> key) {
+        return key.createIdentifier(new EntityWithNullForMissingField(ctx.getFinalEntity(cmd)));
+    }
+
+    @Override
+    public Stream<? extends EntityField<?, ?>> requiredFields(Collection<? extends EntityField<E, ?>> fieldsToUpdate, ChangeOperation op) {
+        return Seq.concat(condition.getFields().stream(), Stream.of(groupingKey.getFields()), Stream.of(groupingKey.getEntityType().getPrimaryKey().getFields()));
```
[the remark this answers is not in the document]
effiban: I'm also not sure I understand why we need the primary keys here - maybe it has to do with cases where a secondary table or parent table is involved.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
galkoren: True. CREATE is indeed the default of the Builder.

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/MaxCountValidator.java
galkoren: Good catch. It's actually a bug. We do need to count even unchanged commands (code was copied from UniquenessValidator)

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -92,6 +94,33 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
...
+        final Field<Integer> countField = count();
+
+        final var queryBuilder = new QueryBuilder<E>(dslContext)
+                .selecting(Seq.<SelectField<?>>seq(aliasedGroupKey.aliasedFields()).append(countField).toList())
+                .from(entityType.getPrimaryTable())
+                .innerJoin(executionPlan.getPaths())
+                .leftJoin(executionPlan.getSecondaryTableRelations())
+                .whereIdsIn(groupingIdentifiers)
+                .withCondition(plCondition.getJooqCondition())
+                .with(q -> q.groupBy(groupKey.getTableFields()));
```
dimagorelik, line 117: @galkoren 
Why not groupBy(groupKey.getTableFields()) instead if custom step (with)?
It is more readable.

effiban (approved):

galkoren:
retest this please

PR #46, opened 2021-08-04
PL-126: Don't audit a created field with a null value

effiban:
In the current audit logic, for a `CREATE` operation - if a field was not populated (has a `null` value), it will still be included in the field changes of the record, with the field name only. However this adds no useful information - since we can omit it altogether, and it will be understood that the value is `null`.

So in this PR I am changing the logic to omit these fields from the record.

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/Triptional.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/Triptional.java
@@ -458,6 +458,40 @@ public class Triptional<T> {
     }
 
     /**
+     * Indicates whether some other object is "equal to" this {@code Triptional}, while comparing them as {@code Optional}-s,
+     * and also using the input function to test equality.<br>
+     * The other object is considered equal if it is also a {@code Triptional}, and <i>also</i> one of the following applies:
+     * <ul>
+     * <li>Both instances have no value present
+     * <li>One of the instances has no value present and the other is {@code null}
+     * <li>Both instances have a value present (possibly {@code null}),
+     * and when the <i>valueEqualityFunction</i> is applied to both values - it returns {@code true}
+     * </ul>
+     *
+     * @param obj an object to be tested for equality
+     * @param valueEqualityFunction the function to use for testing equality of values, when present
+     * @return {@code true} if the other object is "equal to" this object
+     * otherwise {@code false}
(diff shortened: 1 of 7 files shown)


main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -457,6 +457,40 @@ public boolean equals(final Object obj, final BiFunction<T, T, Boolean> valueEqu
...
+
+        if (!(obj instanceof Triptional)) {
+            return false;
+        }
+
+        //noinspection unchecked
+        final Triptional<T> other = (Triptional<T>) obj;
+
+        return valueEqualityFunction.apply(value, other.value);
+    }
```
effiban, line 492: Thw new function equalsAsOptional() behaves the same as a regular equals (with two args), except for the specific case:
When one of the objects is absent and another is null - equalsAsOptional will return true , while regular equals will return false

effiban:
@asiabratslavsky @larisa-bud @shiranyudi - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
```diff
@@ -49,7 +49,7 @@
                                                                     final AuditedField<E, T> field) {
         final var triptionalCurrentValue = auditFieldValueResolver.resolve(field, currentState);
         final var triptionalFinalValue = auditFieldValueResolver.resolve(field, finalState);
-        return triptionalCurrentValue.equals(triptionalFinalValue, field::valuesEqual);
+        return triptionalCurrentValue.equalsAsOptional(triptionalFinalValue, field::valuesEqual);
```
dimagorelik, line 52: @effiban 
Is this logic for create/update operations?

dimagorelik (approved):

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
effiban: @dimagorelik - it's for all operations. `UPDATE` will continue to behave as before, and `CREATE` has one scenario which will behave differently, like we want it to.

PR #47, opened 2021-08-05
PL-127: Don't audit virtual fields

effiban:
Virtual fields are fields calculated from the regular fields of an entity.
Since the regular fields can be audited and they will include all the necessary information about the changes, it's not necessary to audit virtual fields separately, as they will just add extra "noise" to the audit data. 
Therefore this PR disables them.

Should we receive a request in the future, we will add an optional toggle to enable this - but at the moment it doesn't seem likely.


effiban:
@asiabratslavsky @larisa-bud @shiranyudi - FYI

dimagorelik (approved):

PR #49, opened 2021-08-15
PL-128: Support Jooq 3.15.x

effiban:
The support is added by making this the default version, and then separately I will _add_ this version to the list of versions in the release job

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
@@ -77,8 +77,8 @@ public class PersistenceLayerTest {
             {ID_2, TestEnum.Bravo.name(), FIELD2_2_ORIGINAL_VALUE, new Timestamp(Instant.now().toEpochMilli()), "key2", "value2", PARENT_ID_2, COMPLEX_PARENT_ID_2_1, COMPLEX_PARENT_ID_2_2, IGNORABLE_2_ORIGINAL_VALUE},
     };
     private static final Object[][] SECONDARY_DATA = {
-            {ID_1, GOOGLE_URL, ""},
-            {ID_2, DOODLE_URL, ""},
+            {ID_1, GOOGLE_URL},
+            {ID_2, DOODLE_URL},
     };
     private static final Object[][] PARENT_DATA = {
             {PARENT_ID_1, "1000"},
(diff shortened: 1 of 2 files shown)


dimagorelik:
@effiban 
We decided on version 13 if I remember correctly, right?

effiban:
> @effiban
> We decided on version 13 if I remember correctly, right?

@dimagorelik - yes we did, because that is the highest one supported by the **dropwizard-jooq** library. However I looked at this library and realized we can managed without it, and it will simplify the dependencies (see https://github.com/kenshoo/java-dropwizard-gradle-template/pull/81). So if we remove it we can go to the latest one 3.15

integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
```diff
@@ -77,8 +77,8 @@
             {ID_2, TestEnum.Bravo.name(), FIELD2_2_ORIGINAL_VALUE, new Timestamp(Instant.now().toEpochMilli()), "key2", "value2", PARENT_ID_2, COMPLEX_PARENT_ID_2_1, COMPLEX_PARENT_ID_2_2, IGNORABLE_2_ORIGINAL_VALUE},
     };
     private static final Object[][] SECONDARY_DATA = {
-            {ID_1, GOOGLE_URL, ""},
-            {ID_2, DOODLE_URL, ""},
+            {ID_1, GOOGLE_URL},
+            {ID_2, DOODLE_URL},
```
effiban, line 81: The third param is not being used by the batch `insert` on line 128, so I don't know why it's there.
In the previous Jooq versions this didn't cause an error, but it seems that in this new version there is a strict checking that the number of values provided for each row in the batch must match the number of columns exactly - we get an `IndexOutOfBoundsException`

dimagorelik: @effiban 
Have you checked that this could not happen in PL output gen logic?

dimagorelik: @effiban 
I verify it once more and PL does not have this scenario

dimagorelik (approved):

integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
effiban: @dimagorelik - I didn't think about this, I just assumed that if all the tests pass it is ok. But I guess you checked...

PR #52, opened 2021-12-12
Add Byte to built-in String ValueConverters

yagel-n:
In this PR I add the Byte type into the String ValueConverters


effiban (approved):

effiban:
@dimagorelik - FYI

PR #55, opened 2021-12-29
Additional documentation for FieldsCombinationValidator and  FieldValidator

RomanShostakSkai:
After some time of using PL validators we still find it confusing what every method of PL validator is about.
This is an attempt to improve javadoc on them.

@kenshoo/ecom-walmart

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
@@ -18,7 +18,7 @@ import java.util.stream.Stream;
 public interface FieldsCombinationValidator<E extends EntityType<E>> extends ChangeValidator {
 
     /**
-     * @return the list of fields whose combination is being validated
+     * @return the list of fields whose combination is being validated; May contain fields of the current entity only.
      */
     Stream<EntityField<E, ?>> validatedFields();
 
@@ -33,14 +33,14 @@ public interface FieldsCombinationValidator<E extends EntityType<E>> extends Cha
     ValidationError validate(FieldsValueMap<E> fieldsValueMap);
 
     /**
-     * @return a list of fields to fetch.
+     * @return a list of fields to fetch. May contain fields of parent entities only
      */
     default Stream<EntityField<?, ?>> fetchFields() {
         return Stream.of();
     }
 
     /**
-     * @return Predicate when should validate fields.
+     * @return Predicate when should validate fields. It is used together with fetchFields(), so only parent fields can be referenced here.
      */
     default Predicate<CurrentEntityState> validateWhen() {
         return e -> true;
--- a/main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
@@ -19,7 +19,7 @@ public interface FieldValidator<E extends EntityType<E>, T> extends ChangeValida
     EntityField<E, T> validatedField();
 
     /**
-     * Validates the value.
+     * Validates the new field value.
      *
      * @param fieldValue new value of the field
      * @return validation error or <code>null</code> if none


HannaSeraya (approved):

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
```diff
@@ -18,7 +18,7 @@
 public interface FieldsCombinationValidator<E extends EntityType<E>> extends ChangeValidator {
 
     /**
-     * @return the list of fields whose combination is being validated
+     * @return the list of fields whose combination is being validated; Can contains only current entity fields.
```
ilya0508, line 21: **contain**

ilya0508: What happens if developer puts to the stream fields from other entities? I think we need to explain it here

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
```diff
@@ -33,14 +33,14 @@
     ValidationError validate(FieldsValueMap<E> fieldsValueMap);
 
     /**
-     * @return a list of fields to fetch.
+     * @return a list of fields to fetch. Can contain only parent entities fields.
```
ilya0508, line 36: If this is true, I suggest to rename the method as it is very confusing. @dimagorelik WDYT? I know this is a big refactoring, but... 

Also, what happens if there fields from current entity placed to this stream by developer?

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
```diff
@@ -33,14 +33,14 @@
...
      */
     default Stream<EntityField<?, ?>> fetchFields() {
         return Stream.of();
     }
 
     /**
-     * @return Predicate when should validate fields.
+     * @return Predicate when should validate fields. It works with fetchFields(), so Only parent fields can be used here.
```
ilya0508, line 43: can we rephrase it to
`It works in pair with fetchFields(), only fetchFields() fields can be used by predicate.`

What happens if predicate uses some field not present in fetchFields()?

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
dimagorelik: @RomanShostakSkai 
The developer couldn't do it here because the field belongs to generic E

dimagorelik (approved):

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
effiban: I would phrase it like this:
 "It is used together with `fetchFields()`, so only parent fields can be referenced here."

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
effiban: > **contain**

:+1: , I would rephrase it like this:
`May contain fields of the current entity only.`

RomanShostakSkai: @effiban I have changed as you proposed.

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
RomanShostakSkai: @effiban I have changed as you proposed.

main/src/main/java/com/kenshoo/pl/entity/spi/FieldsCombinationValidator.java
RomanShostakSkai: @effiban I have changed as you proposed.

PR #57, opened 2022-01-10
DEVINF-124: Allow overriding the audit field value formatter

effiban:
In Skai we enabled the auditing for a certain entity, and we saw that the default formatting of audited field values is not sufficient since it is not backwards-compatible.
The default uses the PL string value converter of the entity, but we have a different logic that existed before.

To handle this, I am extending the mechanism by allowing the PL client to override the default formatting.
This can be accomplished by implementing a new method in `AuditExtensions`, which is used in the `@Audited` annotation.
The new method returns a value formatter which (if specified) will override the default logic above.

In more detail:
1) Added `AuditExtension.fieldValueFormatter()` which allows the override.
2) Extracted a new hierarchy of classes implementing `AuditFieldValueFormatter`, which replaces a private method in `AuditFieldValueResolver` .
The main logic is implemented in `CompositeAuditFieldValueFormatter`. It will first search for a custom formatter in an `AuditExtensions` if the entity has one, otherwise it will fallback to the default logic which has been extracted to `DefaultAuditFieldValueFormatter`
2) Since we now have two classes examining the `AuditExtensions` annotation, I extracted a shared class for them called `AuditExtensionsExtractor`

**Still TBD - integration tests**


main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditExtensionsExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.annotation.audit.Audited;
+import com.kenshoo.pl.entity.spi.audit.AuditExtensions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Optional;
+
+public class AuditExtensionsExtractor {
```
effiban, line 12: This was extracted from `ExternalMandatoryFieldsExtractor`, and is shared with the new class `CompositeAuditFieldValueFormatter`

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldValueResolver.java
```diff
@@ -21,17 +32,6 @@
...
         return resolve(auditedField, entity)
-            .map(value -> valueToString(auditedField, value));
-    }
-
-    private <T> String valueToString(final AuditedField<?, T> auditedField,
-                                     final T value) {
-        return Optional.ofNullable(auditedField.getStringValueConverter())
-                       .map(converter -> converter.convertTo(value))
-                       .orElse(String.valueOf(value));
-    }
```
effiban, line 32: This logic was extracted to the new hierarchy of value formatters

alexlebedevdev (approved):

yinong (approved):

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditExtensionsExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public Optional<? extends AuditExtensions> extract(final EntityType<?> entityType) {
+        return Optional.ofNullable(entityType.getClass().getAnnotation(Audited.class))
+                       .map(Audited::extensions)
+                       .flatMap(this::createExtensions);
+    }
+
+    private Optional<AuditExtensions> createExtensions(final Class<? extends AuditExtensions> extensionsClass) {
+        try {
+            return Optional.of(extensionsClass.getDeclaredConstructor().newInstance());
```
yinong, line 26: wow... heavy stuff here...

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditExtensionsExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Optional;
+
+public class AuditExtensionsExtractor {
+
+    private static final Logger logger = LoggerFactory.getLogger(AuditExtensionsExtractor.class);
+
+    static final AuditExtensionsExtractor INSTANCE = new AuditExtensionsExtractor();
+
+    public Optional<? extends AuditExtensions> extract(final EntityType<?> entityType) {
+        return Optional.ofNullable(entityType.getClass().getAnnotation(Audited.class))
```
dimagorelik, line 19: @effiban 
Is it run for each field?

dimagorelik:
@effiban 
Can we add the tests for implementing custom logic of AuditFieldValueFormatter

main/src/main/java/com/kenshoo/pl/entity/spi/audit/DefaultAuditFieldValueFormatter.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This is the default formatter for field values to be included in an AuditRecord.<br>
+ * It formats values using the {@code String} {@link com.kenshoo.pl.entity.ValueConverter} defined for a given field.<br>
+ * If no such converter is defined, the default string representation of the value is returned.<br>
+ *
+ * @see EntityField#getStringValueConverter()
+ */
+public class DefaultAuditFieldValueFormatter implements AuditFieldValueFormatter {
```
dimagorelik, line 17: @effiban 
Please, add private constructor

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditExtensionsExtractor.java
effiban: @dimagorelik - yes it is, I will check how to do it more efficiently

main/src/main/java/com/kenshoo/pl/entity/spi/audit/DefaultAuditFieldValueFormatter.java
effiban: :+1:

effiban:
> @effiban Can we add tests for implementing custom logic of AuditFieldValueFormatter?

They will be in the integration tests I will add soon

effiban:
Closing, as I discussed this with @dimagorelik and we will go with a different approach - to add a parameter to the `@Audited` annotation directly and not as an extension.
This will also allow the formatter to be configured per-field and not just per-entity.

PR #60, opened 2022-01-13
DEVINF-124: Allow overriding the audit field value formatter (take 2)

effiban:
_(this is the second version of https://github.com/kenshoo/persistence-layer/pull/57)_

In Skai we enabled the auditing for a certain entity, and we saw that the default formatting of audited field values is not sufficient since it is not backwards-compatible.
The default uses the PL string value converter of the entity, but we have a different logic that existed before.

To handle this, I am extending the mechanism by allowing the PL client to override the default formatting.
This can be accomplished by setting a new optional parameter `valueFormatter`  in the `@Audited` annotation.
If specified, the value formatter will override the default logic above. It can be placed at either the field level or entity-level. If both have it, the field-level overrides the entity-level.

**In more detail:**

- Added a new interface `AuditFieldValueFormatter`, plus a parameter `valueFormatter` of this type to `@Audited`  
   
- Added a default implementation `DefaultAuditFieldValueFormatter` which has the same logic as we had until now: It formats according to  the `String` `ValueConverter` of the entity (if exists), and if not - falls back to the string representation.  
  
- Added a custom implementation `DBAuditFieldValueFormatter` which formats according to the **DB value**. This value can be different than the entity value, if the field has a custom `ValueConverter`.
 **NOTE that a field can have two converters**: The one we are talking about here is responsible for converting to/from the DB value, and is not the same as the `String` `ValueConverter` mentioned above. Refer to the `AbstractEntityType.field(...)` methods for details.
_This custom impl is the one needed by Skai._  
  
- Refactored the code which generates the audited field data, so that it also resolves and stores the value formatter (before running the flow)  
  
- For **external** fields (see `ExternalAuditedField`), the value formatter has been also added as an optional member. If not specified, it will default to the formatter of the underlying field (in the external entity). If the underlying field is not itself audited (in its own entity) - it will simply default to the string representation.

**Still TBD - a few unit tests and all of the integration tests**
  
  

**Usage examples:**

1) Setting a custom formatter at the entity level:

```java
@Audited(valueFormatter = MyValueFormatter.class)
public class MyEntity extends AbstractType<MyEntity> {
      // ....
}
```

2) Setting a custom formatter at the field level:

```java
@Audited
public class MyEntity extends AbstractType<MyEntity> {
      // ...

      @Audited(valueFormatter = MyValueFormatter.class)
      public static final EntityField<MyEntity, String> NAME = INSTANCE.field(MyTable.INSTANCE.name);
}
```

3) Setting a custom value formatter for an _external_ field:

```java
public class MyAuditExtensions implements AuditExtensions {

    @Override
    public Stream<? extends ExternalAuditedField<?, ?>> externalMandatoryFields() {
        return new ExternalAuditedField.Builder(MyParentEntity.NAME)
                                       .withValueFormatter(new MyValueFormatter())
                                       .build();
    }
}

@Audited(extensions = MyAuditExtensions.class)
public class MyEntity extends AbstractType<MyEntity> {
      // ...
}
```


main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
```diff
@@ -30,6 +32,16 @@
...
 
+    /**
+     * @return the formatter to use for field values when they are audited.
+     * <ul>
+     * <li>If defined at the entity level, will apply to all fields in the entity (unless overriden).
+     * <li>If defined at the field level, will apply to that field while overriding any definition (if exists) at the entity level
+     * <li>If missing at both levels, will default to {@link com.kenshoo.pl.entity.spi.audit.DefaultAuditFieldValueFormatter}
+     * </ul>
+     */
+    Class<? extends AuditFieldValueFormatter> valueFormatter() default MissingAuditFieldValueFormatter.class;
```
effiban, line 43: The default used here is not the default formatter, just a dummy invalid formatter that we use to indicate that nothing has been defined - because the compiler will not allow `null` for an annotation param.
This will indicate for example that the field-level annotation is missing and we should use the entity-level one.
We can't put the _default_ here, because then the above will not work (it will never fall back to the entity level when missing on the field).

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldValueFormatterResolver.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public <E extends EntityType<E>> AuditFieldValueFormatter resolve(final EntityField<E, ?> field) {
+        requireNonNull(field, "field is required");
+
+        return Stream.<Supplier<Optional<AuditFieldValueFormatter>>>of(() -> extractFieldLevelFormatter(field),
+                                                                       () -> extractEntityLevelFormatter(field.getEntityType()))
+                     .map(Supplier::get)
+                     .flatMap(Optional::stream)
+                     .findFirst()
+                     .orElse(defaultFormatter);
```
effiban, line 47: In lines 42-43 there are two `Supplier`-s instead of regular method calls.
This is in order to have them evaluated _lazily_, so that if a field-level formatter is found it won't even try to extract the entity-level one, which is more efficient.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldValueResolver.java
```diff
@@ -19,16 +17,8 @@
...
         return resolve(auditedField, entity)
-            .map(value -> valueToString(auditedField, value));
-    }
-
-    private <T> String valueToString(final AuditedField<?, T> auditedField,
-                                     final T value) {
-        return Optional.ofNullable(auditedField.getStringValueConverter())
-                       .map(converter -> converter.convertTo(value))
-                       .orElse(String.valueOf(value));
+            .map(auditedField::formatValue);
```
effiban, line 21: The fixed logic has been replaced by a call to the formatter, which is encapsulated inside `AuditedField` so we just have to call its method.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldResolver.java
```diff
@@ -40,7 +44,8 @@
 
     private <E extends EntityType<E>> AuditedField<E, ?> toAuditedField(EntityField<E, ?> field) {
         final var auditedFieldBuilder = AuditedField.builder(field)
-                                                    .withName(auditFieldNameResolver.resolve(field));
+                                                    .withName(fieldNameResolver.resolve(field))
+                                                    .withValueFormatter(fieldValueFormatterResolver.resolve(field));
```
effiban, line 48: Similar to what was done with the name until now - we resolve the value formatter at the beginning of the flow and store it inside the `AuditedField`.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/ExternalMandatoryFieldsExtractor.java
```diff
@@ -40,6 +53,12 @@
     private AuditedField<?, ?> toAuditedField(final ExternalAuditedField<?, ?> externalAuditedField) {
         return AuditedField.builder(externalAuditedField.getField())
                            .withName(externalAuditedField.getName())
+                           .withValueFormatter(resolveValueFormatter(externalAuditedField))
                            .build();
     }
+
+    private <E extends EntityType<E>> AuditFieldValueFormatter resolveValueFormatter(final ExternalAuditedField<E, ?> externalAuditedField) {
+        return externalAuditedField.getValueFormatter()
+                                   .orElse(valueFormatterResolver.resolve(externalAuditedField.getField()));
```
effiban, line 62: The formatter for the external field is resolved as follows:
- If a formatter has been specified - use it
- If not, take the one that is defined for the underlying field (in the **external** entity) using the same logic as regular fields: First look at the field-level and then fallback to the entity-level if needed.

main/src/test/java/com/kenshoo/pl/entity/internal/audit/entitytypes/AuditedType.java
```diff
@@ -24,6 +25,8 @@
                                                                                   IdentityValueConverter.getInstance(Double.class),
                                                                                   new DoubleToStringValueConverter(),
                                                                                   (a1, a2) -> abs(a1 - a2) < 0.01);
+    public static final EntityField<AuditedType, Double> AMOUNT3 = INSTANCE.field(MainTable.INSTANCE.amount3,
+                                                                                  new DoubleToIntegerValueConverter());
```
effiban, line 29: This is an example of a field with a custom DB `ValueConverter`

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldValueFormatterResolver.java
```diff
@@ -0,0 +1,76 @@
...
+                       .flatMap(this::extractFormatter);
+    }
+
+    private Optional<AuditFieldValueFormatter> extractEntityLevelFormatter(final EntityType<?> entityType) {
+        return Optional.ofNullable(entityType.getClass().getAnnotation(Audited.class))
+                       .flatMap(this::extractFormatter);
+    }
+
+    private Optional<AuditFieldValueFormatter> extractFormatter(final Audited auditedAnnotation) {
+        return Optional.ofNullable(auditedAnnotation.valueFormatter())
```
dimagorelik, line 61: @effiban 
Please, see the warning here

main/src/main/java/com/kenshoo/pl/entity/spi/audit/DBAuditFieldValueFormatter.java
```diff
@@ -0,0 +1,28 @@
...
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+/**
+ * A formatter for audit field values which converts the value into its DB representation using the
+ * {@link com.kenshoo.pl.entity.ValueConverter} defined for the entity, and finally to a string.<br>
+ * If no such converter is defined, the default string representation of the value is returned.<br>
+ * Note that there could be more than one DB value mapped to a single entity value, and in that case the values will be joined by a semicolon.
+ */
+public class DBAuditFieldValueFormatter implements AuditFieldValueFormatter {
```
dimagorelik, line 16: @effiban 
Can we find a more describing name for this Formatter?

dimagorelik:
@effiban 
Looks very good

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldValueFormatterResolver.java
effiban: @dimagorelik - I don't see any warning

main/src/main/java/com/kenshoo/pl/entity/spi/audit/DBAuditFieldValueFormatter.java
effiban: @dimagorelik - I thought about it, but it was hard to find something short. Any ideas?

dimagorelik: @effiban 
Optional.of(auditedAnnotation.valueFormatter())

effiban: @dimagorelik - I guess you meant to comment in the other place. I see your point, it cannot be `null` - but I didn't get any warning for that.

dimagorelik (approved):

PR #63, opened 2022-02-08
add RequiredAtLeastOneChildValidator

ivank-kenshoo:
add new validator interface


main/src/main/java/com/kenshoo/pl/entity/spi/RequiredAtLeastOneChildValidator.java
```diff
@@ -0,0 +1,27 @@
...
+
+public interface RequiredAtLeastOneChildValidator <E extends EntityType<E>, CHILD extends EntityType<CHILD>> extends ChangesValidator<E>{
+
+    CHILD childType();
+
+    @Override
+    default void validate(Collection<? extends EntityChange<E>> entityChanges, ChangeOperation changeOperation, ChangeContext changeContext) {
+        entityChanges.forEach(entityChange -> {
+            long childrenAmount = entityChange.getChildren(childType()).count();
+            if(childrenAmount == 0){
```
ivank-kenshoo, line 15: @dimagorelik WDYT about making it more general? addg `minAmount()` and rename to `RequiredChildValidator`

main/src/main/java/com/kenshoo/pl/entity/spi/RequiredAtLeastOneChildValidator.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Collection;
+
+/**
+ * A validator that force given a child in command
+ *
+ * @param <E> entity type
+ * @param <CHILD> child entity type
+ */
+public interface RequiredAtLeastOneChildValidator <E extends EntityType<E>, CHILD extends EntityType<CHILD>> extends ChangesValidator<E>{
```
dimagorelik, line 13: @ivank-kenshoo 
I do think we should add this CHILD extends EntityType<CHILD>

dimagorelik, line 13: @ivank-kenshoo @effiban 
Maybe it should be a class and not interface
```
public class RequiredAtLeastOneChildValidator <E extends EntityType<E>> extends ChangesValidator<E>{
      
      private final EntityType<?> childType;

      RequiredAtLeastOneChildValidator(EntityTpe<?> childType) {
          this.childType = childType;
      }
```

dimagorelik (approved):

PR #64, opened 2022-02-09
DEVINF-175: Support for adding transient properties to the mandatory audit fields

effiban:
In Skai we have a requirement to add additional properties to an `AuditRecord`-s mandatory fields, that do not come from the DB.

This is accomplished as follows:

1) Adding support for a new type of member in the entity - a **transient** property. These properties hold values which can be populated in a command, but will not be saved to the DB.
2) Adding a new annotation `@AuditDescription` that can be placed on transient properties. When it exists, the property value will be added to the mandatory fields of the audit record.


main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
```diff
@@ -216,4 +238,20 @@ public void add(MissingChildrenSupplier<? extends EntityType> missingChildrenSup
...
     }
+
+    private <T> void add(final TransientEntityField<E, T> transientField, final T newVal) {
+        if (transientValues == null) {
+            transientValues = new HashMap<>();
+        }
+        transientValues.put(transientField, newVal);
+    }
+
+    private void remove(final TransientEntityField<E, ?> transientField) {
```
dimagorelik, line 249: @effiban 
I think we should not support null value in this API

main/src/main/java/com/kenshoo/pl/entity/TransientEntityField.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Represents an entity field which is <i>transient</i>, meaning it is never persisted to the DB.<br>
+ * Transient fields can be set in commands, and the basic PL flow will ignore them and simply carry them through the flow.<br>
+ * They are useful for cases when a client needs to pass additional information about an entity and consume it using some custom logic, e.g. a custom {@link com.kenshoo.pl.entity.spi.OutputGenerator}.
+ *
+ * @param <E> the entity type
+ * @param <T> the value type
+ */
+@SuppressWarnings("unused")
+public interface TransientEntityField<E extends EntityType<E>, T> {
```
dimagorelik, line 12: @effiban 
I think that we shouldn't call to this Field. Maybe TransientEntityObject?

main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
effiban: I didn't want to  - but since the command is mutable anyway, I thought about a case where somebody might have a complex flow where they first populate it with some default value, and then try to clear it

main/src/main/java/com/kenshoo/pl/entity/TransientEntityField.java
effiban: ok, it can be just `TransientObject`

effiban: ok

effiban:
@yinong - FYI

main/src/test/java/com/kenshoo/pl/entity/AbstractEntityTypeTest.java
```diff
@@ -1,148 +1,213 @@
...
+
+import static com.kenshoo.pl.entity.annotation.IdGeneration.Manual;
+import static com.kenshoo.pl.entity.annotation.IdGeneration.RetrieveAutoGenerated;
+import static java.util.stream.Collectors.toUnmodifiableSet;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertFalse;
+
+public class AbstractEntityTypeTest {
```
effiban, line 22: I messed up the formatting here, will fix in one of the next commits

main/src/main/java/com/kenshoo/pl/entity/internal/EntityTypeReflectionUtil.java
```diff
@@ -92,27 +88,27 @@ private static String toEntityFieldName(String methodName) {
         return HashBiMap.create(map);
     }
 
-    private static <E extends EntityType<E>, T> EntityField<E, T> getEntityFieldInstance(final Field field, final EntityType<E> entityType) {
```
effiban, line 95: This method has been moved to the bottom of the class (to keep privates at the end).

main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -24,6 +27,20 @@ public boolean containsField(EntityField<?, ?> field) {
...
 
+    /**
+     * Fetch the value of the given transient property from this final state, if exists.<br>
+     * The value is fetched from the underlying command only (no transient properties exist in an {@code Entity})
+     *
+     * @param transientProperty the property whose value is to be fetched; required
+     * @param <T> the type of value to be fetched
+     * @return the value, or {@code Optional.empty()} if it does not exist
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
dimagorelik, line 38: @effiban 
Why do we need it here?

dimagorelik: @effiban 
The final entity represents the DB final state

effiban: The code which generates the audit values is in `AuditMandatoryFieldValuesGenerator`. It receives a `FinalEntityState` because that is how we can determine the correct value to be audited for a field.
For transient properties we have different logic - we only need to look at the command, but I would rather not split the logic into 2 generators.
If you think about it, you could say that the logic is the same: The final state first looks for a value in the command, and only if not found - fetch from the current state because command override current state. A transient property is never in the DB, so logically it is the same - you can imagine that we look it up in the DB and always get an absent result.

Since we can't actually look up the transient property in the DB, I did have to add a new method which only looks in the command - but I think it is reasonable.
If you think it is a big issue, I will create a new generator that looks in the command only.

main/src/test/java/com/kenshoo/pl/entity/TestEntityWithTransient.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class TestEntityWithTransient extends AbstractEntityType<TestEntityWithTransient> {
+
+    public static final TestEntityWithTransient INSTANCE = new TestEntityWithTransient();
+
+    @Id
+    public static final EntityField<TestEntityWithTransient, Integer> ID = INSTANCE.field(TestEntityTable.TABLE.id);
+    public static final EntityField<TestEntityWithTransient, String> FIELD_1 = INSTANCE.field(TestEntityTable.TABLE.field_1);
+
+    public static final TransientEntityObject<TestEntityWithTransient, String> TRANSIENT_1 = INSTANCE.transientObject("transient_1");
```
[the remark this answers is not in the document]
effiban: Yes, I thought about this. The only reason for having a type is so that the compiler will force you to use the same type when you try to set or get the value in a command.
We cannot guarantee (even for a regular field, not just transient) that when you get the value it will have the correct type, but at least when you write the relevant you will be aligned to the same type that is declared here.

effiban:
Going with a different approach, as discussed with @dimagorelik

PR #65, opened 2022-02-14
fix error code for RequiredAtLeastOneChildValidator

ivank-kenshoo:
missed the fact that ValidationError should receive errorCode, not error message. this is client responsibility to define error code and handle message on top of it

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/validators/RequiredAtLeastOneChildValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/validators/RequiredAtLeastOneChildValidator.java
@@ -4,6 +4,7 @@ import com.kenshoo.pl.entity.*;
 import com.kenshoo.pl.entity.spi.ChangesValidator;
 
 import java.util.Collection;
+import java.util.Map;
 
 /**
  * A validator that force given a child in command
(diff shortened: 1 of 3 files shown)


main/src/main/java/com/kenshoo/pl/entity/validators/RequiredAtLeastOneChildValidator.java
```diff
@@ -25,7 +27,7 @@ public void validate(Collection<? extends EntityChange<E>> entityChanges, Change
             long childrenAmount = entityChange.getChildren(childType).count();
             if (childrenAmount == 0) {
                 changeContext.addValidationError(entityChange,
-                        new ValidationError(String.format("At least one %s is required.", childType.getName())));
+                        new ValidationError(errorCode));
```
dimagorelik, line 30: @ivank-kenshoo 
We do not need to supply what is child type in the error code, right?

ivank-kenshoo: right. I think because both childType and errorCode in constructor now, better to allow client to set its own errorCode

dimagorelik (approved):

PR #66, opened 2022-02-14
Support EntityChangeDescription transient property for auditing

effiban:


main/src/main/java/com/kenshoo/pl/entity/TransientProperty.java
```diff
@@ -0,0 +1,20 @@
...
+ * The basic PL flow will ignore these properties and simply carry them through the flow.<br>
+ * It can be useful for cases where a client wants to pass additional information to be consumed at the end of the flow, without saving it to the DB.<br>
+ * The first use-case to be implemented within PL is a property holding a custom description for auditing.
+ */
+public interface TransientProperty<T> {
+
+    /**
+     * @return the name of the property
+     */
+    String getName();
```
effiban, line 14: can be removed

effiban: replace this with a description of the property

main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
```diff
@@ -64,6 +67,14 @@ public void set(Collection<EntityField<E, ?>> fields, MultiFieldValueSupplier<E>
         }
     }
 
+    @Override
+    public <T> void set(final TransientProperty<T> transientProperty, final T newValue) {
```
dimagorelik, line 71: @effiban 
I do not like parameter name newValue. Let's call it propertyValue or propertyObject or transientObject

main/src/main/java/com/kenshoo/pl/entity/internal/TransientPropertyImpl.java
```diff
@@ -0,0 +1,50 @@
+package com.kenshoo.pl.entity.internal;
+
+import com.kenshoo.pl.entity.TransientProperty;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import static java.util.Objects.requireNonNull;
+import static org.apache.commons.lang3.Validate.notBlank;
```
dimagorelik, line 8: @effiban 
org.apache.commons.lang3?

main/src/test/java/com/kenshoo/pl/entity/EntityCommandExt.java
```diff
@@ -3,15 +3,19 @@
...
-
 public interface EntityCommandExt<E extends EntityType<E>, C extends EntityCommandExt<E, C>> extends MutableCommand<E> {
 
     default <T> C with(EntityField<E, T> field, T newValue) {
         set(field, newValue);
         return (C) this;
     }
 
+    @SuppressWarnings("unchecked")
+    default <T> C with(final TransientProperty<T> property, final T newValue) {
```
dimagorelik, line 14: @effiban 
See comment above

main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
effiban: ok , will change to `propertyValue`

main/src/main/java/com/kenshoo/pl/entity/internal/TransientPropertyImpl.java
effiban: Yes, why not? We use it anyway for equals / hashCode.
It's better than Guava (which we also have) because it should not cause as many dependency conflicts. I would rather get rid of Guava, if anything..
In addition - we should upgrade to Gradle 7 which gives you the option to properly hide your implementation dependencies, and then we won't have to worry about transitive dependency issues.

main/src/test/java/com/kenshoo/pl/entity/EntityCommandExt.java
effiban: :+1:

effiban:
I ran basic performance tests with the IntelliJ profiler for the memory consumed by commands (without running PL - just creating command objects). Results:

- An empty properties map holds **64 bytes**, so for 10K commands it adds **640K** compared to no map at all
- A properties map with one value of 20 chars holds **148 bytes**, so for 10K commands adds **840K** compared to the empty map, and **1.48M** compared to no map at all

PR #67, opened 2022-02-27
DEVINF-239: Fix `FinalEntityState` getter for fields with asymmetric ValueConverters

effiban:
A PL field can define a `ValueConverter` containing custom logic for transforming the field value from the PL representation to the DB representation, and vice versa.
In most cases these two transformations will be _symmetric_, meaning that if you transform in both directions you will get back the _same_ value that you started with.

However this is not enforced anywhere and we want to leave some flexibility for cases where the PL type and DB type are not fully compatible.
**For example**: 
Suppose your application handles numeric values with a scale of up to 2 fractional digits. The values are retrieved from some 3rd-party which provides values in that scale, and the values are saved to the DB to a column which also supports the same scale. After some time the 3rd-party API changes and now values are provided with 3 fractional digits.
Now suppose that you cannot or don't want to change the DB column definition - so to support this, you would need to change your `ValueConverter` so that:
- It rounds 3 digits to 2 digits for saving to the DB
- It converts 2 digits back to 3 digits (padding with zero) when loading from DB.
So in this case if the PL command got a `BigDecimal` value of **12.119**, it would be saved to the DB as **12.12** and loaded back as **12.120**

This causes a bug in at least 2 features of PL: **auditing** and **purger** . Both of these features work by comparing the value of some field in a command to the current value of the same field in the DB. If the values are equal in their DB representation but the field converter is asymmetric, then in the example above we could have **12.119** in the cmd and **12.120** in the current value (after loading) - so the comparison would return false. 
The comparison is controlled by a "valueEqualityFunction" that can also be overriden for the field, so if this function were to be defined properly - it would "cancel out" the discrepancy.
However I think it's too much to assume that a PL user will properly coordinate between them, and it's extremely difficult if not impossible to enforce this.

In this PR, I am fixing the issue for **auditing** which is where we discovered it in Skai.
The auditing flow performs the comparison mentioned above using `FinalEntityState.get()` - and so the fix will be there, and the new logic will go as follows: 
- Fetch value from command (if exists)
- Convert to DB value
- Convert back to PL value
- Now compare this value to the current value

The back-and-forth conversion ensures that we will end up with a properly "normalized" value to compare (even when the "valueEqualityFunction" is not properly designed).


main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -29,4 +29,11 @@ public boolean containsField(EntityField<?, ?> field) {
         throw new UnsupportedOperationException("Final state of children is unknown at this point");
     }
 
+    private <E extends EntityType<E>, T> T getFromEntityChange(EntityField<E, T> field) {
+        final var dbAdapter = field.getDbAdapter();
```
dimagorelik, line 33: @effiban 
Let's speak

PR #70, opened 2022-05-24
inverse value converter

ivank-kenshoo:
I am faced with the need to convert Long db field to have String in the code. 
We already have `CommonTypesStringConverter` that converts String to custom type.
`StringValueCommonConverter` is opposite implementation, converts custom type to String.

Before pushing tests, @dimagorelik @effiban WDYT?


effiban:
> I am faced with the need to convert Long db field to have String in the code. We already have `CommonTypesStringConverter` that converts String to custom type. `StringValueCommonConverter` is opposite implementation, converts custom type to String.
> 
> Before pushing tests, @dimagorelik @effiban WDYT?

@ivank-kenshoo - I think there is a simpler solution. `CommonTypesStringConverter` already converts in both directions, the only problem is the order of the generic params. So your new converter can simply call the existing one using the opposite methods, in other words:
`StringValueCommonConverter.convertFrom()` will call `CommonTypesStringConverter.convertTo()`, and vice versa. 
If there are any missing conversion functions, simply add them to the existing maps in `CommonTypesStringConverter`.
If you do this I think maybe a better name for your class would be `ReverseCommonTypesStringConverter` or something like that.

ivank-kenshoo:
@effiban great idea! please review

effiban:
@ivank-kenshoo - looks good, please go ahead and add tests

PR #73, opened 2022-05-26
DEVINF-366: Support auditing of an `EntityType` with no id

effiban:
In Skai we have a use-case for auditing of a one-to-many connection, in which the child entity has a primary key which is a _composite_ of two fields. The first of these is a foreign key to the parent id. 
Therefore, there is no single field that can be marked with `@Id`.
In addition , even if we were to choose the first one to be the `@Id`, it wouldn't work for `CREATE` because of how PL currently handles foreign keys during create.

We decided that we don't wish to change the behavior of `@Id` - it should be used only to mark a **single numeric** primary key, which is the best practice for all tables (a table could also have a unique "logical" key which is a composite). 
However since many tables in practice don't have such a primary key, especially child tables - we decided to support this case by making the `entityId` **optional** for auditing. It is assumed that in such a case, there will be a unique logical key and its fields will be marked as `@Audited(trigger = ALWAYS)` so that the entity can still be identified uniquely in the auditing storage through the mandatory fields.

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateTwoLevelsManualOrNoIdTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateTwoLevelsManualOrNoIdTest.java
@@ -11,7 +11,6 @@ import com.kenshoo.pl.entity.internal.audit.entitytypes.*;
 import org.jooq.DSLContext;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import java.util.List;
@@ -103,7 +102,6 @@ public class AuditForCreateTwoLevelsManualOrNoIdTest {
                 hasCreatedFieldRecord(AuditedChildManualIdType.NAME, CHILD_NAME))));
     }
 
-    @Ignore
     @Test
     public void oneParentWithManualId_OneChildWithoutId_ShouldCreateRecordsForParentAndChild() {
         final var childCmd = new CreateAuditedChildWithoutIdOfParentManualIdCommand()
@@ -131,12 +129,11 @@ public class AuditForCreateTwoLevelsManualOrNoIdTest {
                 hasCreatedFieldRecord(AuditedManualIdType.NAME, PARENT_NAME)));
 
         assertThat(auditRecord, hasChildRecordThat(allOf(hasEntityType(AuditedChildWithoutIdOfParentManualIdType.INSTANCE.getName()),
+                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentWithoutIdType.NAME, CHILD_NAME),
                 hasOperator(CREATE),
-                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentManualIdType.PARENT_ID, String.valueOf(PARENT_ID)),
                 hasCreatedFieldRecord(AuditedChildWithoutIdOfParentManualIdType.NAME, CHILD_NAME))));
     }
 
-    @Ignore
     @Test
     public void oneParentWithoutId_OneChildWithoutId_ShouldCreateRecordsForParentAndChild() {
         final var childCmd = new CreateAuditedChildWithoutIdOfParentWithoutIdCommand()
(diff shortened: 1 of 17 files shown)


integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateTwoLevelsManualOrNoIdTest.java
```diff
@@ -131,12 +129,11 @@ public void oneParentWithManualId_OneChildWithoutId_ShouldCreateRecordsForParent
                 hasCreatedFieldRecord(AuditedManualIdType.NAME, PARENT_NAME)));
 
         assertThat(auditRecord, hasChildRecordThat(allOf(hasEntityType(AuditedChildWithoutIdOfParentManualIdType.INSTANCE.getName()),
+                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentWithoutIdType.NAME, CHILD_NAME),
                 hasOperator(CREATE),
-                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentManualIdType.PARENT_ID, String.valueOf(PARENT_ID)),
```
effiban, line 135: We still have a limitation where we cannot retrieve the parentId during `CREATE` (for the same reason we couldn't retrieve the field marked as `@Id` in the original Skai bug), so instead of checking for this one, I added a check for the other mandatory field which was missing before.
As we discussed offline, it is ok to have only the child mandatory field in the auditing record because it is recursive and we can still identify the parent.

integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateTwoLevelsManualOrNoIdTest.java
```diff
@@ -164,7 +161,7 @@ public void oneParentWithoutId_OneChildWithoutId_ShouldCreateRecordsForParentAnd
 
         assertThat(auditRecord, hasChildRecordThat(allOf(hasEntityType(AuditedChildWithoutIdOfParentWithoutIdType.INSTANCE.getName()),
                 hasOperator(CREATE),
-                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentWithoutIdType.PARENT_NAME, String.valueOf(PARENT_NAME)),
+                hasMandatoryFieldValue(AuditedChildWithoutIdOfParentWithoutIdType.NAME, CHILD_NAME),
```
effiban, line 164: as above

main/src/test/java/com/kenshoo/pl/entity/internal/audit/AuditRequiredFieldsCalculatorTest.java
```diff
@@ -26,340 +27,481 @@ public class AuditRequiredFieldsCalculatorTest {
     private static final ChangeOperation OPERATOR = ChangeOperation.UPDATE;
 
     @Test
-    public void requiredFields_FieldSetHasIdOnly_FieldsToChangeAreDifferent_ShouldReturnIdOnly() {
-        final AuditedEntityType<AuditedAutoIncIdType> auditedEntityType =
-            builder(AuditedAutoIncIdType.ID).build();
+    public void requiredFields_TypeHasIdOnly_ToChangeAreDifferent_ShouldReturnIdOnly() {
+        final var auditedEntityType = builder(AuditedAutoIncIdType.INSTANCE).build();
```
effiban, line 31: In this test class I made many style changes to shorten lines and names. 
The important changes (new test) start at line 350 (view in 'Split' mode)

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRequiredFieldsCalculator.java
```diff
@@ -26,13 +25,12 @@ public AuditRequiredFieldsCalculator(final AuditedEntityType<E> auditedEntityTyp
...
-        final Seq<? extends EntityField<E, ?>> intersectedChangeFields =
-            seq(fieldsToChange).filter(onChangeFields::contains);
+        final var intersectedChangeFields = seq(fieldsToChange).filter(onChangeFields::contains);
 
-        return Seq.<EntityField<?, ?>>of(auditedEntityType.getIdField())
-            .append(auditedEntityType.getUnderlyingMandatoryFields())
-            .append(intersectedChangeFields);
+        return Seq.<EntityField<?, ?>>seq(auditedEntityType.getUnderlyingMandatoryFields())
+                .append(intersectedChangeFields)
+                .append(auditedEntityType.getIdField());
```
effiban, line 34: This is the actual change (the rest is style)

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
```diff
@@ -42,10 +42,18 @@ public String getEntityType() {
         return entityType;
     }
 
+    /**
+     * @deprecated replaced by {@link #safeGetEntityId()} since the entity id is no longer mandatory
+     */
+    @Deprecated
```
dimagorelik, line 48: @effiban 
Why do you need the old property? Does it break something?

effiban: I couldn't do an exact search in Github for this method, but clients are definitely using `AuditRecord` , so to be safe I am keeping the old method

dimagorelik (approved):

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
effiban: I will remove the old method, as discussed offline

PR #77, opened 2022-09-11
Allow when condition on field validator

dimagorelik:
Allow when condition on field validator.
The change will allow to check condition on Parent entities before running validation

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
@@ -29,7 +29,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
-import static com.kenshoo.pl.entity.EntityForTest.FIELD1;
 import static com.kenshoo.pl.entity.EntityForTest.URL;
 import static com.kenshoo.pl.entity.spi.FieldValueSupplier.fromOldValue;
 import static com.kenshoo.pl.entity.spi.FieldValueSupplier.fromValues;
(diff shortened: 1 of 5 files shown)


larisa-bud (approved):

main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
```diff
@@ -25,4 +29,18 @@
      * @return validation error or <code>null</code> if none
      */
     ValidationError validate(T fieldValue);
+
+    /**
+     * @return a list of fields to fetch. May contain fields of parent entities only
+     */
+    default Stream<EntityField<?, ?>> fetchFields() {
```
effiban, line 36: You probably copied this from `ImmutableFieldValidator` - but I think a better name would be `parentFieldsToFetch()`.
I remember discussing this name already some time ago.....

integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
```diff
@@ -1234,6 +1252,19 @@ public ValidationError validate(Integer fieldValue) {
         }
     }
 
+    private static class TestField2WihParentConditionValidator extends TestField2Validator {
```
effiban, line 1255: typo: WthParent -> W**i**thParent

integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
```diff
@@ -573,6 +575,22 @@ public void validField2InvalidValueForCreateValidation() {
...
+
+        EntityChangeCompositeValidator<EntityForTest> changesCompositeValidator = new EntityChangeCompositeValidator<>();
+        changesCompositeValidator.register(new TestField2WihParentConditionValidator());
+        ChangeFlowConfig<EntityForTest> flowConfig = changeFlowConfig().withValidator(changesCompositeValidator).build();
+
+        UpdateResult<EntityForTest, EntityForTest.Key> updateResult = persistenceLayer.update(commands, flowConfig);
+
+        Collection<ValidationError> validationErrors = updateResult.getErrors(commands.get(0));
+        assertThat(validationErrors.iterator().next().getErrorCode(), is(FIELD2_ERROR));
+    }
```
effiban, line 592: Please also add a test for the case where the condition is not met (where the parent id doesn't match so the validator is not applied). 
I see you did that in a unit test, but would be nice to have this here as well if it doesn't take too long.

main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
dimagorelik: @effiban 
I didn't copy from ImmutableFieldValidator.
In all our API fetchFields means parent's fields.
We should fix all APIs id different initiative

effiban: ah ok, I forgot that

effiban (approved):

PR #87, opened 2024-01-29
Support new field operators

dimagorelik:
Support new field operators:
1. IsNotNull - generate <> SQL condition
2. eq between two fields - generate f1.eq(f2)

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
@@ -134,6 +134,17 @@ public class PLContextSelectTest {
     }
 
     @Test
+    public void selectNothingIsNotNull() {
+        final List<CurrentEntityState> entities = plContext.select(TestParentEntityType.FIELD1)
+                .from(TestParentEntityType.INSTANCE)
+                .where(TestParentEntityType.FIELD2.isNotNull())
+                .fetch();
+        assertThat("Incorrect number of entities fetched: ",
+                entities.size(), is(1));
+
+    }
+
+    @Test
     public void selectFromChildAndParent() {
         final List<CurrentEntityState> entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1, TestParentEntityType.FIELD1)
                                                .from(TestEntityType.INSTANCE)
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
@@ -53,9 +53,9 @@ public class PLContextSelectTest {
             tablesCreated = true;
         }
         dslContext.insertInto(parent_table)
-                  .columns(parent_table.id, parent_table.field1)
-                  .values(1, "ParentAlpha")
-                  .values(2, "ParentBravo")
+                  .columns(parent_table.id, parent_table.field1, parent_table.field2)
+                  .values(1, "ParentAlpha", null)
+                  .values(2, "ParentBravo", "ParentBravo")
                   .execute();
 
         dslContext.insertInto(parent_sec_table)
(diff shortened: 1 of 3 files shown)


main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -36,6 +36,18 @@ default PLCondition eq(T value) {
...
+    default PLCondition eq(EntityField<?, T> otherField) {
+        if (isVirtual()) {
+            throw new UnsupportedOperationException("The equals operation is unsupported for virtual fields");
+        }
+
+        @SuppressWarnings("unchecked")
+        final TableField<Record, T> tableField = (TableField<Record, T>)getDbAdapter().getFirstTableField();
+        @SuppressWarnings("unchecked")
+        final TableField<Record, T> otherTableField = (TableField<Record, T>)otherField.getDbAdapter().getFirstTableField();
+        return new PLCondition(tableField.eq(otherTableField), entity -> entity.safeGet(this).equalsValue(entity.safeGet(otherField).get()), this, otherField);
```
effiban, line 48: Please use `equals()` instead of `equalsValue()` - you are comparing two Triptionals here,  so it needs to be a regular comparison.

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
```diff
@@ -96,12 +96,38 @@ public void selectFromSingleEntity() {
...
+                .where(TestParentEntityType.FIELD1.eq(TestParentEntityType.FIELD2))
+                .fetch();
+        assertThat("Incorrect number of entities fetched: ",
+                entities.size(), is(1));
+
+        assertThat(entities.get(0),
+                hasFieldValues(fieldValue(TestParentEntityType.ID, 2),
+                        fieldValue(TestParentEntityType.FIELD1, "ParentBravo"),
+                        fieldValue(TestParentEntityType.FIELD2, "ParentBravo")));
+    }
```
effiban, line 112: Please add a test for the case where both are `NULL`

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
```diff
@@ -96,12 +96,38 @@ public void selectFromSingleEntity() {
...
+
+    @Test
+    public void selectIsNotNull() {
+        final List<CurrentEntityState> entities = plContext.select(TestParentEntityType.FIELD1)
+                .from(TestParentEntityType.INSTANCE)
+                .where(TestParentEntityType.FIELD1.isNotNull())
+                .fetch();
         assertThat("Incorrect number of entities fetched: ",
                 entities.size(), is(2));
```
effiban, line 132: Please add a test for the opposite case where the field is `NULL`

main/src/test/java/com/kenshoo/pl/entity/EntityFieldConditionsTest.java
```diff
@@ -39,6 +39,27 @@ public void testFieldInValuesInPostFetchCondition() {
...
+
+        assertThat(TestEntityType.NAME1.eq(TestEntityType.NAME2).getPostFetchCondition().test(entity), is(true));
+    }
+
+    @Test
+    public void testTwoFieldNotEqualInPostFetchCondition() {
+        final var entity = createEntityWith(TestEntityType.NAME1, "myName", TestEntityType.NAME2, "anotherName");
+
+        assertThat(TestEntityType.NAME1.eq(TestEntityType.NAME2).getPostFetchCondition().test(entity), is(false));
+    }
```
effiban, line 61: please add more tests for checking equals when:
- exactly one is NULL
- both are NULL

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
dimagorelik: @effiban 
Do you mean 
entity.safeGet(this).equals(entity.safeGet(otherField).get())?

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
dimagorelik: @effiban 
This case is not supported by MySql

effiban: @dimagorelik - sorry, you're right. I was mixing up the two cases, when we translate to SQL and when we do a comparison in Java only.

effiban (approved):

effiban (approved):

PR #88, opened 2024-02-05
DRAFT - Support comparable entity fields and DSL

effiban:


main/src/main/java/com/kenshoo/pl/entity/AbstractEntityType.java
```diff
@@ -67,6 +64,52 @@ protected <T> EntityField<E, T> field(EntityFieldDbAdapter<T> dbAdapter, ValueCo
         return addField(new EntityFieldImpl<>(this, dbAdapter, stringValueConverter, Objects::equals));
     }
 
+    protected <T extends Comparable<T>> EntityField<E, T> comparableField(final TableField<Record, T> tableField) {
+        return comparableField(tableField, IdentityValueConverter.getInstance(tableField.getType()));
+    }
```
effiban, line 69: In Java we cannot overload with the same name and params when the only difference is the generic type - so i had to choose a different name at least for some of the methods.
To remain consistent, I decided it is better to use the same name for all of them in any case.

main/src/main/java/com/kenshoo/pl/entity/ComparableEntityField.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Objects;
+import java.util.function.Predicate;
+
+public interface ComparableEntityField<E extends EntityType<E>, T extends Comparable<T>> extends EntityField<E, T> {
+
+    default PLCondition greaterThan(T value) {
+        if (isVirtual()) {
+            throw new UnsupportedOperationException("The greaterThan operation is unsupported for virtual fields");
+        }
```
effiban, line 14: As we discussed, we will probably change this condition so it is only unsupported for virtual fields with more than one underlying field.

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -280,10 +280,11 @@ private Predicate<? super EntityChange<?>> withOperator(ChangeOperation op) {
         return cmd -> op == cmd.getChangeOperation();
     }
 
-    private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
+    private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, ?> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        //noinspection unchecked
+        cmd.set((EntityField<ROOT, Object>)idField,  currentState.get(idField));
```
effiban, line 287: The original casting was moved to here which is more precise, because only here we are actually mutating the value

main/src/test/java/com/kenshoo/pl/entity/TestEntity.java
```diff
@@ -14,6 +21,9 @@ public class TestEntity extends AbstractEntityType<TestEntity> {
     public static final PrototypedEntityField<TestEntity, Integer> FIELD_3 = INSTANCE.prototypedField(TestDataFieldPrototype.FIELD_3, TestEntityTable.TABLE.field_3);
     public static final EntityField<TestEntity, String> SECONDARY_FIELD_1 = INSTANCE.field(SecondaryTable.TABLE.secondary_field_1);
 
+    public static final EntityField<TestEntity, BigDecimal> CUSTOM_COMPARABLE_FIELD =
+            INSTANCE.field(TestEntityTable.TABLE.field_4).comparedBy(comparing(n -> n.setScale(2, DOWN)));
```
effiban, line 25: A dummy example of how to configure a custom comparator - not used in any tests

main/src/main/java/com/kenshoo/pl/entity/internal/AbstractEntityField.java
```diff
@@ -0,0 +1,148 @@
...
+    private Comparator<T> comparatorOf(T value) {
+        return Optional.ofNullable(valueComparator)
+                .orElseGet(() -> {
+                    if (value instanceof Comparable<?>) {
+                        //noinspection unchecked
+                        return (Comparator<T>) Comparator.naturalOrder();
+                    }
+                    throw new UnsupportedOperationException(String.format("The field [%s] is not comparable for the value [%s]", this, value));
+                });
+    }
```
effiban, line 97: This implementation is designed to add support for comparison operations for all fields as much as possible.
- If the field is comparable:
     - It will use the natural comparator by default.
     - If the user overrode this, it will use the specific one
- If the field is not comparable - an exception will be thrown

main/src/main/java/com/kenshoo/pl/entity/MutableEntityField.java
```diff
@@ -0,0 +1,8 @@
+package com.kenshoo.pl.entity;
+
+import java.util.Comparator;
+
+public interface MutableEntityField<E extends EntityType<E>, T> extends EntityField<E, T> {
+
+    EntityField<E, T> comparedBy(final Comparator<T> valueComparator);
+}
```
effiban, line 8: This interface is needed for the DSL to work properly - after a field has been created with `field()` we need to be able to optionally add a comparator. On the other hand we don't weant this to be part of the regular `EntityField` interface used in many places that should be effectively 'immutable'

main/src/main/java/com/kenshoo/pl/entity/internal/EntityFieldImpl.java
```diff
@@ -1,61 +1,48 @@
...
-    }
+        public Builder(final EntityType<E> entityType) {
+            super(entityType);
+        }
 
-    @Override
-    @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
-    public boolean equals(Object o) {
-        return (this == o);
-    }
```
effiban, line 55: Removed this because it is the default equals implementation (if I'm not mistaken), and I didn't put it in the new base class either

main/src/main/java/com/kenshoo/pl/entity/internal/EntityFieldImpl.java
```diff
@@ -1,61 +1,48 @@
...
+        protected Builder<E, T> self() {
+            return this;
+        }
 
-    @Override
-    public int hashCode() {
-        return System.identityHashCode(this);
```
effiban, line 59: as above, I don't know why this was overriden

main/src/main/java/com/kenshoo/pl/entity/MutableEntityField.java
dimagorelik: @effiban 
Let's call this interface ComparableEntityField

effiban: The prolem is that it is only comparable **after** you call the method - so if you use this name, then every field we have today will implement this interface, which might be confusing.
Need to think of something similar that means "optionally comparable"

PR #89, opened 2024-02-18
Split PLCondition into DB and post-fetch conditions

effiban:


main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -27,55 +31,75 @@ default boolean isVirtual() {
...
+    }
+
+    private PLPostFetchCondition postFetchCompareWith(final Predicate<Triptional<T>> predicate) {
+        return new PLPostFetchCondition(entity -> predicate.test(entity.safeGet(this)), this);
+    }
+
+    private TableField<Record, Object> extractFirstTableField() {
         if (isVirtual()) {
-            throw new UnsupportedOperationException("The equals operation is unsupported for virtual fields");
+            throw new UnsupportedOperationException("PLConditions cannot be built for virtual fields");
```
effiban, line 96: As discussed this should be replaced later on with a condition that checks the number of fields - but right now we don't have a method returning number of fields, and it would probably not be efficient

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
```diff
@@ -96,21 +96,6 @@ public void selectFromSingleEntity() {
...
-                .where(TestParentEntityType.FIELD1.eq(TestParentEntityType.FIELD2))
-                .fetch();
-        assertThat("Incorrect number of entities fetched: ",
-                entities.size(), is(1));
-
-        assertThat(entities.get(0),
-                hasFieldValues(fieldValue(TestParentEntityType.ID, 2),
-                        fieldValue(TestParentEntityType.FIELD1, "ParentBravo"),
-                        fieldValue(TestParentEntityType.FIELD2, "ParentBravo")));
-    }
```
effiban, line 112: Remove because this method is changed to work only for post-fetch

main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -27,55 +31,75 @@ default boolean isVirtual() {
...
         }
-        @SuppressWarnings("unchecked")
-        final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
-        return new PLCondition(tableField.isNotNull(), entity -> entity.safeGet(this).isNotNull(), this);
+        //noinspection unchecked
+        return (TableField<Record, Object>)getDbAdapter().getFirstTableField();
+    }
+
+    private boolean valuesEqual(final Triptional<T> thisTriptional, final T otherValue) {
+        return thisTriptional.matches(thisValue -> valuesEqual(thisValue, otherValue));
```
effiban, line 103: This method calls the **public** `valuesEqual()` and that way, ensures that the equality function will be used and not the default equals (this fixes a bug we had).

main/src/main/java/com/kenshoo/pl/entity/PLPostFetchCondition.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static java.util.Objects.requireNonNull;
+
+public class PLPostFetchCondition implements PLBaseCondition<PLPostFetchCondition> {
```
dimagorelik, line 12: @effiban 
Let's call it PLEntityCondition

main/src/main/java/com/kenshoo/pl/entity/PLCondition.java
```diff
@@ -1,76 +1,62 @@
...
 import org.jooq.Condition;
 import org.jooq.impl.DSL;
 
 import java.util.Set;
-import java.util.function.Predicate;
 
 import static java.util.Objects.requireNonNull;
 
-public class PLCondition {
+public class PLCondition implements PLBaseCondition<PLCondition> {
```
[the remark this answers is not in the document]
effiban: I thought about it, but wasn't sure if we want to refactor this name - but I like the name

main/src/main/java/com/kenshoo/pl/entity/PLPostFetchCondition.java
effiban: ok, but I'm actually not sure it will be clear

main/src/main/java/com/kenshoo/pl/entity/PLBaseCondition.java
```diff
@@ -0,0 +1,14 @@
+package com.kenshoo.pl.entity;
+
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+public interface PLBaseCondition<C extends PLBaseCondition<C>> {
```
dimagorelik, line 7: @effiban 
Why do we need a common interface?

PR #90, opened 2024-02-19
Enhance basic validator interfaces

dimagorelik:
Fix interface definitions of predicates for next validators FieldValidator and RequiredFieldValidator

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
@@ -1,11 +1,8 @@
 package com.kenshoo.pl.entity.spi;
 
-import com.kenshoo.pl.entity.CurrentEntityState;
-import com.kenshoo.pl.entity.EntityField;
-import com.kenshoo.pl.entity.EntityType;
-import com.kenshoo.pl.entity.ValidationError;
+import com.kenshoo.pl.entity.*;
 
-import java.util.function.Predicate;
+import java.util.function.BiPredicate;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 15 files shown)


main/src/test/java/com/kenshoo/pl/entity/internal/validators/RequiredFieldValidationAdapterTest.java
```diff
@@ -94,7 +94,7 @@ public void when_require_value_and_predicate_false_return_null() {
         CreateEntityCommand<TestEntity> entityChange = new CreateEntityCommand<>(TestEntity.INSTANCE);
         entityChange.set(TestEntity.FIELD_1, (String)null);
 
-        when(validator.requireWhen()).thenReturn(currentState -> false);
+        when(validator.requireWhen()).thenReturn((currentState, change) -> false);
```
dimagorelik, line 97: @effiban 
For required field validation, the current state and the final state are the same. Maybe anyway we should receive here only the final state?

main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
```diff
@@ -38,9 +35,9 @@ public interface FieldValidator<E extends EntityType<E>, T> extends ChangeValida
     }
 
     /**
-     * @return Predicate when should validate fields. It is used together with fetchFields(), so only parent fields can be referenced here.
+     * @return Predicate when should validate fields. It is used together with fetchFields()
      */
-    default Predicate<CurrentEntityState> validateWhen() {
-        return e -> true;
+    default BiPredicate<CurrentEntityState, EntityChange<E>> validateWhen() {
```
dimagorelik, line 40: @effiban 
The field value validator does not know for which operation it's applied UPDATE or CREATE. It means that the relevant value for the predicate condition could be in the current state or change. I think a much better and more user-friendly interface for Predicate is FinalEntityState. What do you think?

effiban: @dimagorelik - that makes sense, but you can also make the opposite argument:
What if the validator wants to do use some different logic for the field in the current state compared to the same field in the change - then he won't be able to this at all if he only has the final state.
Anyway it is a question of what type of conditions are the most common. I guess whatever you think is best

PR #91, opened 2024-02-19
Enhance basic validator interfaces

dimagorelik:
Enhance interfaces for next validators FieldValidator and RequiredFieldValidator

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
@@ -1,5 +1,13 @@
 package com.kenshoo.pl.entity;
 
+import com.kenshoo.jooq.SelectQueryExtender;
+
+/**
+ * The current state of an entity in the database.
+ * This is the state of the entity before any changes are applied and contains the current values of the entity's fields
+ * which are requested by consumers.
+ *
+ */
 public interface CurrentEntityState extends Entity {
 
     CurrentEntityState EMPTY = new CurrentEntityState() {
--- a/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
@@ -1,11 +1,9 @@
 package com.kenshoo.pl.entity;
 
-import com.kenshoo.jooq.SelectQueryExtender;
-
 /**
  * The current state of an entity in the database.
  * This is the state of the entity before any changes are applied and contains the current values of the entity's fields
- * which are requested by consumers.
+ * which are requested by consumers like validators, enrichers and etc.
  *
  */
 public interface CurrentEntityState extends Entity {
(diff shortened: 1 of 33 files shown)


main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
```diff
@@ -1,5 +1,13 @@
 package com.kenshoo.pl.entity;
 
+import com.kenshoo.jooq.SelectQueryExtender;
+
+/**
```
dimagorelik, line 5: @effiban 
Please, validate the description

main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -2,12 +2,22 @@
 
 import java.util.List;
 
+/**
```
dimagorelik, line 5: @effiban 
Please, validate the description

main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
effiban: Looks good 👍 
One comment - I would add a short explanation or link to another class to explain what a "consumer" is.

main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
effiban: Looks good but I would use the future tense because the changes have not been applied yet.
Something like this:
"This **will be** the state of the entity in the DB after all changes **have been** applied ...."

main/src/main/java/com/kenshoo/pl/entity/spi/FieldValidator.java
```diff
@@ -38,9 +35,10 @@ public interface FieldValidator<E extends EntityType<E>, T> extends ChangeValida
     }
 
     /**
-     * @return Predicate when should validate fields. It is used together with fetchFields(), so only parent fields can be referenced here.
+     * @return Predicate when should validate fields. It is used together with fetchFields().
```
effiban, line 38: I suggest changing the phrasing a bit and adding some information:

```@return a predicate indicating when the field should be validated. It will be evaluated together with {@link #fetchFields()}, which means that all of the fields appearing in the predicate must also be included in the fields to fetch.```

File: main/src/main/java/com/kenshoo/pl/entity/spi/RequiredFieldValidator.java [hunk 4]
```diff
@@ -26,9 +26,10 @@ public interface RequiredFieldValidator<E extends EntityType<E>, T> extends Chan
     String getErrorCode();
 
     /**
-     * @return Predicate when should validate field.
+     * @return Predicate when should validate fields. It is used together with fetchFields().
```
effiban, line 29: See previous comment

main/src/main/java/com/kenshoo/pl/entity/spi/ImmutableFieldValidator.java
```diff
@@ -33,7 +33,8 @@ public interface ImmutableFieldValidator<E extends EntityType<E>, T> extends Cha
     }
 
     /**
-     * @return Predicate when should validate field.
+     * @return Predicate when should validate field. It is used together with fetchFields().
```
effiban, line 36: as above

effiban (approved):

PR #93, opened 2024-02-22
Simplify fields combination validator interface

dimagorelik:
Simplify fields combination validator interface


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/FieldsCombinationCreator.java
```diff
@@ -15,7 +15,7 @@ public class FieldsCombinationCreator {
 
     public static <E extends EntityType<E>> Collection<? extends FieldsValueMap<E>> create(final Collection<? extends EntityChange<E>> entityChanges, final Collection<EntityField<E, ?>> fields, ChangeOperation changeOperation, ChangeContext changeContext) {
```
dimagorelik, line 16: @effiban 
This utility class is not used by the framework. Maybe we can remove it from here and recreate it in the client code?

main/src/main/java/com/kenshoo/pl/entity/internal/FieldsCombination.java
```diff
@@ -12,13 +12,18 @@
 
 import static java.util.stream.Collectors.toSet;
 
-public class ResultingFieldsCombination<E extends EntityType<E>> implements FieldsValueMap<E> {
+/**
+ * The final state of fields in the database.
```
effiban, line 16:
```suggestion
 * The final state of a certain combination of fields in the database.
```

main/src/main/java/com/kenshoo/pl/entity/internal/FieldsCombination.java
```diff
@@ -12,13 +12,18 @@
...
+ * This will be the state of fields after all changes have been applied on the current state and contains the final values of the entity's fields.
+ *
+ */
+public class FieldsCombination<E extends EntityType<E>> implements FieldsValueMap<E> {
```
effiban, line 20: How about `FinalFieldsCombination` ?

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/FieldsCombinationCreator.java
effiban: @dimagorelik - not sure, it looks like this is a helper that makes things easier for the client so we wouldn't want to move it there - but I haven't looked at the usages yet.

main/src/main/java/com/kenshoo/pl/entity/internal/FieldsCombination.java
dimagorelik: @effiban 
In this case, I prefer FieldsCombination

effiban (approved):

PR #94, opened 2024-02-29
Enhance prototype validator interfaces

dimagorelik:


main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldComplexValidator.java
```diff
@@ -32,5 +30,14 @@ public interface PrototypeFieldComplexValidator<T> extends ChangeValidator {
...
      */
-    Stream<EntityField<?, ?>> fetchFields();
+    default Stream<EntityField<?, ?>> fetchFields() { return Stream.of(); }
+
+    /**
+     * The predicate is evaluated on the final state of the entity See {@link FinalEntityState}.
+     * @return a predicate indicating when the field should be validated. It will be evaluated together with {@link #fetchFields()},
+     * which means that all the parent fields appearing in the predicate must also be included in the fields to fetch.
+     */
+    default Predicate<CurrentEntityState> validateWhen() {
```
dimagorelik, line 40: @effiban 
The new validateWhen API allows checking conditions only for parent entities because the prototype interface does not belong to any specific entity. This is the same common logic for all prototype interfaces.

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldComplexValidator.java
```diff
@@ -32,5 +30,14 @@ public interface PrototypeFieldComplexValidator<T> extends ChangeValidator {
     /**
      * @return a list of fields to fetch. Can contain only parent entities fields.
      */
-    Stream<EntityField<?, ?>> fetchFields();
+    default Stream<EntityField<?, ?>> fetchFields() { return Stream.of(); }
```
effiban, line 33: I think we should rename this to `parentFieldsToFetch`.
If you think that's too specific then at least `fieldsToFetch` as in other places, which is better than `fetchFields` (because that one is a verb indicating an action)

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldValidator.java
```diff
@@ -22,4 +24,18 @@ public interface PrototypeFieldValidator<T> extends ChangeValidator {
      * @return validation error or <code>null</code> if none
      */
     ValidationError validate(T value);
+
+    /**
+     * @return a list of fields to fetch. Can contain only parent entities fields.
+     */
+    default Stream<EntityField<?, ?>> fetchFields() { return Stream.of(); }
```
effiban, line 31: Same comment as above

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldsCombinationValidator.java
```diff
@@ -25,4 +25,18 @@ public interface PrototypeFieldsCombinationValidator extends ChangeValidator {
      * @return validation error if there is or <code>null</code> if OK
      */
     ValidationError validate(PrototypeFieldsCombination<?> fieldsCombination);
+
+    /**
+     * @return a list of fields to fetch. Can contain only parent entities fields.
+     */
+    default Stream<EntityField<?, ?>> fetchFields() { return Stream.of(); }
```
effiban, line 32: as above

main/src/test/java/com/kenshoo/pl/entity/internal/validators/PrototypeFieldValidationAdapterTest.java
```diff
@@ -50,16 +55,14 @@ public void testSupportedOperation() {
     }
 
     @Test
-    public void testFetchFieldsInUpdate() {
-        Stream<? extends EntityField<?, ?>> fields = adapter.fieldsToFetch();
-        assertFalse("Do not fetch field", fields.findFirst().isPresent());
+    public void testFetchFields() {
+        when(validator.fetchFields()).thenReturn(Stream.of(field1, field2));
+        Collection<? extends EntityField<?, ?>> fields = adapter.fieldsToFetch().collect(toList());;
```
effiban, line 60: Please use `toUnmodifiableList` and also remove the extra `;` at the end

main/src/test/java/com/kenshoo/pl/entity/internal/validators/PrototypeFieldValidationAdapterTest.java
```diff
@@ -50,16 +55,14 @@ public void testSupportedOperation() {
...
+        assertEquals("Do not fetch field", fields.size(), 2);
+        assertTrue("Fetch field1", fields.contains(field1));
+        assertTrue("Fetch field2", fields.contains(field2));
```
effiban, line 63: These error messages are not clear, and also I prefer to use `assertThat`, so I would do it like this:
```
assertThat("Incorrect number of fields to fetch", fields.size(), is(2));
assertThat("Missing first field to fetch", fields, contains(field1));
assertThat("Missing second field to fetch", fields, contains(field2));
```

main/src/test/java/com/kenshoo/pl/entity/internal/validators/PrototypeFieldsCombinationValidationAdapterTest.java
```diff
@@ -59,21 +65,26 @@ public void testSupportedOperation() {
...
 
     @Test
-    public void testFetchFieldsInCreate() {
+    public void testFetchFieldsWithParentFields() {
+        when(validator.fetchFields()).thenReturn(Stream.of(field1, field2));
         Collection<? extends EntityField<?, ?>> fields = adapter.fieldsToFetch().collect(toSet());
         assertTrue("Fetch field1", fields.contains(TestEntity.FIELD_1));
         assertTrue("Fetch field2", fields.contains(TestEntity.FIELD_2));
+        assertTrue("Fetch field2", fields.contains(field1));
+        assertTrue("Fetch field2", fields.contains(field2));
```
effiban, line 82: As above

main/src/test/java/com/kenshoo/pl/entity/internal/validators/PrototypeFieldsCombinationValidationAdapterTest.java
```diff
@@ -85,16 +96,28 @@ public void testValidateForCreate() {
...
+    @Test
+    public void testSkipValidateValue() {
+        when(validator.validateWhen()).thenReturn(value -> false);
+        adapter.validate(entityChange, currentState, finalState);
+        verify(validator,  never()).validate(argThat(fieldCombination -> {
+            assertEquals("Field1", fieldCombination.get(TestDataFieldPrototype.FIELD_1), STRING_VALUE1);
+            assertEquals("Field2", fieldCombination.get(TestDataFieldPrototype.FIELD_2), STRING_VALUE2);
+            return true;
+        }));
+    }
```
effiban, line 116: Here you are validating the "never" case' so you don't need to check all the fields inside, you can just do this:
```
 verify(validator,  never()).validate(any());
```

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldComplexValidator.java
dimagorelik: @effiban 
Maybe better ancestorsFields?

effiban: @dimagorelik - that's also good

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldComplexValidator.java
```diff
@@ -30,7 +28,16 @@ public interface PrototypeFieldComplexValidator<T> extends ChangeValidator {
...
-     * @return a list of fields to fetch. Can contain only parent entities fields.
+     * @return a list of fields to fetch. Can contain only ancestor entities fields.
      */
-    Stream<EntityField<?, ?>> fetchFields();
+    default Stream<EntityField<?, ?>> ancestorsFields() { return Stream.of(); }
+
+    /**
+     * The predicate is evaluated on the final state of the entity See {@link FinalEntityState}.
+     * @return a predicate indicating when the field should be validated. It will be evaluated together with {@link #ancestorsFields()},
+     * which means that all the ancestor fields appearing in the predicate must also be included in the fields to fetch.
```
effiban, line 38:
```suggestion
     * which means that all the ancestor fields appearing in the predicate must also be included in the response of {@link #ancestorsFields()}.
```

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldsCombinationValidator.java
```diff
@@ -25,4 +25,18 @@ public interface PrototypeFieldsCombinationValidator extends ChangeValidator {
      * @return validation error if there is or <code>null</code> if OK
      */
     ValidationError validate(PrototypeFieldsCombination<?> fieldsCombination);
+
+    /**
+     * @return a list of fields to fetch. Can contain only ancestor entities fields.
```
effiban, line 30: as above

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldsCombinationValidator.java
```diff
@@ -25,4 +25,18 @@ public interface PrototypeFieldsCombinationValidator extends ChangeValidator {
...
+     */
+    default Stream<EntityField<?, ?>> ancestorsFields() { return Stream.of(); }
+
+    /**
+     * The predicate is evaluated on the final state of the entity See {@link FinalEntityState}.
+     * @return a predicate indicating when the field should be validated. It will be evaluated together with {@link #ancestorsFields()},
+     * which means that all the ancestors fields appearing in the predicate must also be included in the fields to fetch.
```
effiban, line 37: as above

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldValidator.java
```diff
@@ -22,4 +24,18 @@ public interface PrototypeFieldValidator<T> extends ChangeValidator {
      * @return validation error or <code>null</code> if none
      */
     ValidationError validate(T value);
+
+    /**
+     * @return a list of fields to fetch. Can contain only ancestor entities fields.
```
effiban, line 29: as above

main/src/main/java/com/kenshoo/pl/entity/spi/PrototypeFieldValidator.java
```diff
@@ -22,4 +24,18 @@ public interface PrototypeFieldValidator<T> extends ChangeValidator {
...
+     */
+    default Stream<EntityField<?, ?>> ancestorsFields() { return Stream.of(); }
+
+    /**
+     * The predicate is evaluated on the final state of the entity See {@link FinalEntityState}.
+     * @return a predicate indicating when the field should be validated. It will be evaluated together with {@link #ancestorsFields()},
+     * which means that all the ancestors fields appearing in the predicate must also be included in the fields to fetch.
```
effiban, line 36: as above

effiban (approved):

PR #99, opened 2024-12-23
Set idField only when it exists to avoid throwing an exception when performing a dummy insert or upsert.

andriikoziichuk:
[SEARCH-173084](https://kenshoo.atlassian.net/browse/SEARCH-173084)

PL | fix review mode for create and upsert commands

I faced an issue where the dummy upsert or insert did not work. 
To fix it, we decided to set the primary field only when it exists.

`java.lang.RuntimeException: java.lang.IllegalArgumentException: Field ID of entity "sov keyword" is not fetched
`

[SEARCH-173084]: https://kenshoo.atlassian.net/browse/SEARCH-173084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
@@ -283,7 +283,7 @@ public class PersistenceLayer<ROOT extends EntityType<ROOT>> {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        cmd.set(idField,  currentState.safeGet(idField).asOptional().orElse(null));
     }
 
     private DSLContext dslContext() {
--- a/main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
@@ -283,7 +283,9 @@ public class PersistenceLayer<ROOT extends EntityType<ROOT>> {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.safeGet(idField).asOptional().orElse(null));
+        if (currentState.containsField(idField)) {
+            cmd.set(idField, currentState.get(idField));
+        }
     }
 
     private DSLContext dslContext() {
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PersistenceLayerTest.java
@@ -1062,7 +1062,7 @@ public class PersistenceLayerTest {
         assertThat(insertOnDuplicateUpdateResult.getStats().getAffectedRowsOf(mainTable.getName()).getUpdated(), is(0));
 
         boolean result = insertOnDuplicateUpdateResult.getChangeResults().stream()
-                .allMatch(changeResult -> changeResult.getCommand().get(ID) == null);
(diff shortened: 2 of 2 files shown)


main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -283,7 +283,7 @@ private Predicate<? super EntityChange<?>> withOperator(ChangeOperation op) {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        cmd.set(idField,  currentState.safeGet(idField).asOptional().orElse(null));
```
dimagorelik, line 286: @andriikoziichuk 
I prefer to skip set id field does not exist

andriikoziichuk: @dimagorelik changed. WDYT?

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -283,7 +283,7 @@ private Predicate<? super EntityChange<?>> withOperator(ChangeOperation op) {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        currentState.safeGet(idField).ifNotNull(id -> cmd.set(idField, id));
```
dimagorelik, line 286: @effiban 
Can you also review this change?

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -283,7 +283,9 @@ private Predicate<? super EntityChange<?>> withOperator(ChangeOperation op) {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        if (currentState.containsField(idField)) {
```
dimagorelik, line 286: @effiban 
What should we do when an id field does not exist in preview mode because this uses the db autoincrement trigger? 
Should we return something or skip it?

effiban: I guess we could try to generate the **expected** next id, using the autoincrement sequence or something like that. Don't remember if this is possible

andriikoziichuk: Do we really need it? In preview mode we interested only in the validation results

effiban: @andriikoziichuk - I think it's better to have an id, to maintain a uniform creation logic shared between the 'preview' (dummy) mode and regular mode. You can think of it as maintaining the same "contract" for the client in both cases. It could also help to distinguish between the results (although not sure where that really matters).  

In terms of performance, the performance should not be any slower than the real case, and I think we can live with it. However it's not absolutely critical

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
effiban: Looks good to me

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
andriikoziichuk: @effiban 
Can we take this solution as temporary and create a Jira item for your team to handle appropriately at a later date? 
Now the “preview” mode does not work for create and upsert commands

effiban: @andriikoziichuk - yeah, that's fine with me

effiban (approved):
Please create the Jira for the long-term solution

andriikoziichuk:
The story https://kenshoo.atlassian.net/browse/SEARCH-173096

andriikoziichuk:
thanks! I found the article.

